PR #47, opened 2017-07-24
feat(autoscaler): Added autoscaler to yaml

### LINKED ISSUE #31: figure out how to scale down cluster
right now, when I want to scale up the cluster, I just up the min and max in the aws autoscaling group for the kubernete worker nodes.
However I am not sure what's the safe way to scale it down while still making sure that all the deployments still maintain the desired replica numbers?

### LINKED ISSUE #46: fix server_name in nginx conf & domain name in revproxy_setup.sh for ssl cert

zflamig:
Resolves #31 
Fixes #46

### DIFF
--- a/destroy.bash
+++ b/destroy.bash
@@ -9,16 +9,8 @@ OUTPUT_DIR=`ls -d *_output | head -1`
 echo "Working with $OUTPUT_DIR"
 VPCNAME=`echo $OUTPUT_DIR | cut -d_ -f1`
 
-# Destroy services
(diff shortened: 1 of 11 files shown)


destroy.bash
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+cd tf_files
+LOGIN_NODE=`grep -A20 "aws_eip.login" terraform.tfstate | grep "public_ip" | head -1 | sed 's/[ \",]//g' | cut -d: -f2`
+
+echo "Working with Login Node: $LOGIN_NODE"
+
```
zflamig: https://coreos.com/kubernetes/docs/latest/kubernetes-on-aws-destroy.html the notes seem to suggest no?

PR #59, opened 2017-07-31
feat(portal): get schema before running nginx

philloooo:
required for https://github.com/uc-cdis/data-portal/pull/91
r? @thanh-nguyen-dang

### DIFF
--- a/tf_files/configs/00configmap.yaml
+++ b/tf_files/configs/00configmap.yaml
@@ -1,0 +1,8 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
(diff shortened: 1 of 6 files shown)


tf_files/configs/00configmap.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: global
+  namespace: default
+data:
+  environment: ${vpc_name}
```
thanh-nguyen-dang: where do you use this key? `environment`

PR #61, opened 2017-08-04
feat(proxy): use squid as NAT instance

### LINKED ISSUE #55: replace internet gateway with nginx proxy for k8s

philloooo:
resolve #55 
- removed NAT gateway
- moved tf_variables and tfstate to `~/.creds/{vpc_name}` so that I can use cloud-automation with different vpcs locally
- used envsubst instead of sed for tf_variables templating
- some minor fixes

tested in OCC with `pr65` vpc name :p

r? @zflamig

### DIFF
--- a/pack_and_terraform.bash
+++ b/pack_and_terraform.bash
@@ -1,10 +1,18 @@
 #!/bin/bash
 # Right now this script guides you through the process
 # of running Packer & Terraform
-
+unamestr=`uname`
+if [[ "$unamestr" == 'Darwin' ]]; then
+    export PATH=${PATH}:/usr/local/opt/gettext/bin
+    if [[ ! -x "$(command -v envsubst)" ]]; then
+	echo "need envsubst to run this script, please install gettext"
+	exit 1
+    fi
+fi
 if [ -z "$AWS_ACCESS_KEY" ]; then
 	read -p "Enter your AWS ACCESS key: " AWS_ACCESS_KEY
 fi
+
 if [ -z "$AWS_SECRET_KEY" ]; then
 	read -p "Enter your AWS SECRET key: " AWS_SECRET_KEY
 fi
@@ -17,7 +25,7 @@ echo "Got key for $GITHUB: $SSHKEY"
 
 read -n 1 -p "Build packer images (y/n)? " BUILDPACKER
 [ -z "$BUILDPACKER" ] && answer="No"
-echo 
+echo
 
 if echo "$BUILDPACKER" | grep -iq "^y"; then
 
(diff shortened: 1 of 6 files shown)


pack_and_terraform.bash
```diff
@@ -1,10 +1,17 @@
 #!/bin/bash
 # Right now this script guides you through the process
 # of running Packer & Terraform
-
+unamestr=`uname`
+if [[ "$unamestr" == 'Darwin' ]]; then
+    export PATH=${PATH}:/usr/local/opt/gettext/bin
+    if [[ ! -x "$(command -v envsubst)" ]]; then
+        brew install gettext
```
zflamig: Can we not do brew (I don't have it)? Lets either compile from source or just error out if missing and give the user a couple of options.

pack_and_terraform.bash
```diff
@@ -79,78 +86,79 @@ if echo "$RUNTF" | grep -iq "^y"; then
...
+    export SSHKEY=$SSHKEY
+    export ADDSSHKEY=$SSHKEY
+    export BUCKET=$BUCKET
+    export CLIENTSECRET=$CLIENTSECRET
+    export CLIENTID=$CLIENTID
+    export HMAC=$HMAC
+    export USERAPIDBPASS=$USERAPIDBPASS
+    export INDEXDDBPASS=$INDEXDDBPASS
+    export GDCAPIDBPASS=$GDCAPIDBPASS
+    export INDEXD=$INDEXD
```
zflamig: Can we pass these just to envsubst so we don't pollute the env variables globally?

setup_kube.bash
```diff
@@ -1,13 +1,18 @@
...
 cd tf_files
-LOGIN_NODE=`grep -A20 "aws_eip.login" terraform.tfstate | grep "public_ip" | head -1 | sed 's/[ \",]//g' | cut -d: -f2`
+LOGIN_NODE=`grep -A20 "aws_eip.login" $creds_dir/terraform.tfstate | grep "public_ip" | head -1 | sed 's/[ \",]//g' | cut -d: -f2`
 
 echo "Working with Login Node: $LOGIN_NODE"
 
-OUTPUT_DIR=`ls -d *_output | head -1`
 echo "Working with $OUTPUT_DIR"
-VPCNAME=`echo $OUTPUT_DIR | cut -d_ -f1`
+OUTPUT_DIR=${VPC_NAME}_output
```
zflamig: This should be above the echo

pack_and_terraform.bash
philloooo: sure, why do you care as we only have envsubst and terraform in the last though? it won't pollute anything outside of the script

PR #67, opened 2017-08-09
docs(architecture): updated high level doc

philloooo:
updated minimal docs thanks to @allisonheath 's diagram
r? @zflamig


philloooo:
hmmm I thought through the VPC endpoint we allow both ways?

zflamig:
Correct. But from the desktops to the S3 is functionally 1-way

philloooo:
oh you are talking about that arrow.
I think finally all projects will have some portion of released data in s3 that allow download from outside. bpa might never do that in case they don't want to share any results

PR #68, opened 2017-08-09
feat(kube-elb): remove external revproxy

### LINKED ISSUE #57: replace external revproxy with kube revproxy service and ELB

philloooo:
resolve #57 
- require certificate manager to be configured before using terraform
- move k8s related configs to a separate kube.tf in case we want to support only standing up VPC for some projects later
- minor fixes
Tested in test environment.
The cert is created for test.bionimbus.org that can be reused later
r? @zflamig


kube/services/gdcapi/gdcapi-deploy.yaml
```diff
@@ -1,9 +1,9 @@
-apiVersion: extensions/v1beta1 
+apiVersion: extensions/v1beta1
 kind: Deployment
 metadata:
   name: gdcapi-deployment
 spec:
-  replicas: 1 
+  replicas: 2
```
zflamig: Why default 2 now?

setup_kube.bash
```diff
@@ -14,7 +14,7 @@ OUTPUT_DIR=${VPC_NAME}_output
 echo "Working with $OUTPUT_DIR"
 
 if [ ! -f $OUTPUT_DIR/cdis-devservices-secret.yml ]; then
-	echo "Please provide cdis-devservices-secret.yml in $OUTPUT_DIR before proceeding"
+	echo "Please provide cdis-devservices-secret.yml in tf_files/$OUTPUT_DIR before proceeding"
 	exit
 fi
```
zflamig: Do we need to remove the proxy stuff from here now?

tf_files/kube.tf
```diff
@@ -0,0 +1,235 @@
+
```
zflamig: We should have a style for organizing these so that like resource blocks end up near each other

kube/services/gdcapi/gdcapi-deploy.yaml
philloooo: 2 is the minimum number to make rolling new versions no downtime

zflamig: Can we specify a surge instead? https://tachingchen.com/blog/Kubernetes-Rolling-Update-with-Deployment/

PR #71, opened 2017-08-14
feat(s3): push terraform.state to s3 bucket

thanh-nguyen-dang:
resolve issue #56 by:
- adding the backend configuration in `cloud.tf` file.
- adding the file to save the credentials of aws s3.
- initializing the backend by running command: `terraform init -backend-config="terrafom.tfvars"`
r @philloooo @zflamig

### DIFF
--- a/pack_and_terraform.bash
+++ b/pack_and_terraform.bash
@@ -106,6 +106,13 @@ if echo "$RUNTF" | grep -iq "^y"; then
     creds_dir=$HOME/.creds/$VPC_NAME
     mkdir -p $creds_dir
 
+	if [ -z "$AWS_S3_ACCESS_KEY" ]; then
+		read -p "Enter your access key to S3 bucket: " AWS_S3_ACCESS_KEY
+	fi
+
+	if [ -z "$AWS_S3_SECRET_KEY" ]; then
+		read -p "Enter your secret key to S3 bucket: " AWS_S3_SECRET_KEY
+	fi
 
 	if [ -z "$SOURCE_AMI" ]; then
 		read -p "Enter your base ami: " SOURCE_AMI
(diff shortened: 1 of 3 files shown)


tf_files/cloud.tf
```diff
@@ -1,3 +1,11 @@
+terraform {
+    backend "s3" {
+        bucket = "cdis-terraform-state"
+        key    = "cdis-vpc/terraform.tfstate"
```
philloooo: change this to `cdis-terraform-{vpc_name}/terraform.tfstate`?

thanh-nguyen-dang: the backend configuration cannot contain interpolations. So, we need to input the whole key or hardcode it. Like I did in `terraform.tfvars` for `access_key` and `secret_key`

tf_files/cloud.tf
```diff
@@ -1,3 +1,10 @@
+terraform {
+    backend "s3" {
+        bucket = "cdis-terraform-state"
+        region = "us-east-1"
```
zflamig: We should abstract out the region so that the state bucket can live anywhere

pack_and_terraform.bash
```diff
@@ -140,19 +147,25 @@ if echo "$RUNTF" | grep -iq "^y"; then
 	fi
 
 	export LC_CTYPE=C
-	HMAC=`cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 32 | head -n 1 | base64`
+	HMAC=`cat /dev/urandom | env LC_CTYPE=C | tr -dc 'a-zA-Z0-9' | fold -w 32 | head -n 1 | base64`
```
zflamig: This won't work because you are passing the output of `env` to `tr` so you don't get an entropy

pack_and_terraform.bash
```diff
@@ -106,6 +106,13 @@ if echo "$RUNTF" | grep -iq "^y"; then
     creds_dir=$HOME/.creds/$VPC_NAME
     mkdir -p $creds_dir
 
+	if [ -z "$AWS_S3_ACCESS_KEY" ]; then
+		read -p "Enter your access key to S3 bucket: " AWS_S3_ACCESS_KEY
```
zflamig: Can we adding something about "for saving terraform state" here in the prompt so it is more clear?

thanh-nguyen-dang:
resolve the conflicts! thanks for your review again @zflamig

zflamig:
@rudyardrichter Can you test this too?

philloooo:
resolved by #73

PR #72, opened 2017-08-15
feat(gdcapi-secret): add secret key

philloooo:
added secret key to work with https://github.com/uc-cdis/gdcapi/pull/39
r? @rudyardrichter

### DIFF
--- a/pack_and_terraform.bash
+++ b/pack_and_terraform.bash
@@ -146,16 +146,20 @@ if echo "$RUNTF" | grep -iq "^y"; then
 	fi
 
 	export LC_CTYPE=C
-	HMAC=`cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 32 | head -n 1 | base64`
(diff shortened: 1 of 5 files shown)


pack_and_terraform.bash
```diff
@@ -149,6 +149,10 @@ if echo "$RUNTF" | grep -iq "^y"; then
 	HMAC=`cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 32 | head -n 1 | base64`
 	echo "Your HMAC encryption key is: $HMAC"
 
+
+	GDCAPI_SECRET=`cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 50 | head -n 1`
```
rudyardrichter: I think this should be:
```sh
base64 /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 50 | head -n 1
```
(at least on my machine this way works but `cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 50 | head -n 1` does not)

PR #75, opened 2017-08-16
fix(script): clean up pack_and_terraform.bash

rudyardrichter:
- Print errors raised by packer
- Exit on packer failure
- Factor repeated tasks into functions
- Clean up indentation

A few opinionated changes—the indentation was inconsistent in some places so I made it all 4 spaces.

r? @philloooo @zflamig


rudyardrichter:
Since the AWS secret key is allowed to contain slashes, the sed for setting variables in the packer variables file also needs to be changed to use a different delimiter.

PR #78, opened 2017-08-17
fix/revert

thanh-nguyen-dang:
revert the wrong commit
@philloooo


philloooo:
closing it as it's not reverting only the backend change and we are merging the fix soon

PR #93, opened 2017-10-11
feat(migration): Supports migrating databases

### LINKED ISSUE #35: encrypt etcd
https://kubernetes.io/docs/tasks/administer-cluster/encrypt-data/

### LINKED ISSUE #63: support vpc migration
support creating database from snapshots

zflamig:
WIP
Resolves #63 
Resolves #35


philloooo:
- all variables are changed to match ones in tf_variables
- misc fixes unrelated to this pr
  - removed redundant & unused nat eip 
  - ignore host checking because scripts try to ssh to kube.internal.io through various login vms and that will just break them
  - fixed s3 backend setup
- added one if condition in `pack_and_terraform` to ask user if it's a migration, if so, fetch creds from previous vpc's kube provisioner node and source those vars in current shell.
- the database names had to be changed to remove the prefix - because when we use snapshots we inherit the names and renaming dbs manually will be disruptive(db endpoints will change)

tested with deploying vpc `pr93` then -> `pr93blue` ->`pr93green`, then tested with a clean `pr93new` for regression
this will make our future regression pretty painless, although becase I changed the database name, migrating existing VPCs needs to manually change the database names new VPC is up
r? @zflamig

zflamig:
lgtm

PR #96, opened 2017-10-30
fix(workdir): run terraform in state dir

frickjack:
some tweaks to support running terraform in
a state directory that contains .tfvars files
relevant for a particular terraform deployment

so we do something like this:
* cd deployment-name/
* terraform init -backend-config ./backend.tfvars
* terraform plan -var-file ./config.tfvars
   -out plan.terraform
   ~/Code/cloud-automation/tf_files/aws/v1/

Also added Jenkins/ folder:
Jenkins
├── Docker
│   └── Jenkins
│       └── Dockerfile
├── Pipelines
│   ├── Backup
│   │   └── Jenkinsfile
│   ├── Terraform
│   │   └── Jenkinsfile
│   └── TerraformPlanOnly
│       └── Jenkinsfile
├── README.md
└── Stacks
    └── Jenkins
        └── docker-compose.yml


tf_files/aws/variables.tf
```diff
@@ -37,9 +42,22 @@ variable "kube_ssh_key" {
    kubernete nodes, Example:
    '- ssh-rsa XXXX\n - ssh-rsa XXX' */
 variable "kube_additional_keys" {
-    default = ""
+    default = <<EOB
```
philloooo: let's not add those keys as default? admins don't need to ssh onto pods

tf_files/aws/terraform.tfvars
```diff
@@ -1,5 +0,0 @@
-access_key = "$AWS_S3_ACCESS_KEY"
```
philloooo: let's keep this file, this was accidentally unused but fixed in https://github.com/uc-cdis/cloud-automation/pull/93/commits/1943eb611b69e5dd168c3e3f7b632b33fe07e3a1

pack_and_terraform_aws.bash
```diff
@@ -147,21 +147,33 @@ if echo "$RUNTF" | grep -iq "^y"; then
     if [ -z "$AWS_S3_ACCESS_KEY" ]; then
         read -p "Enter your access key to S3 bucket for saving terraform state: " AWS_S3_ACCESS_KEY
     fi
+    # Terraform variable: 
+    #  https://www.terraform.io/docs/configuration/environment-variables.html#tf_var_name
+    export TF_VAR_access_key="${AWS_S3_ACCESS_KEY}"
```
philloooo: let's not do this? should be fixed [here](https://github.com/uc-cdis/cloud-automation/pull/93/commits/1943eb611b69e5dd168c3e3f7b632b33fe07e3a1#diff-40a95469d103a81a11be31908b3fc11eR264) to use those creds for `init` but exporting them as tf vars are error prone because AWS_ACCESS_KEY might be different from AWS_S3_ACCESS_KEY, and we might not use the creds to access s3 for terraform apply

Jenkins/Pipelines/TerraformPlanOnly/Jenkinsfile
```diff
@@ -0,0 +1,45 @@
...
+
+pipeline {
+  agent any
+
+  stages {
+    stage('FetchCode'){
+      steps {
+        checkout scm
+        dir('Secrets') {
+            sh 'aws s3 cp s3://cdis-terraform-state/planx-pla.net/v1/config.tfvars config.tfvars'
```
philloooo: what does version mean here?

Jenkins/Pipelines/Terraform/Jenkinsfile
```diff
@@ -0,0 +1,56 @@
+#!groovy
+
+pipeline {
+  agent any
+
+  stages {
+    stage('FetchCode'){
+      steps {
+        git url: "https://github.com/uc-cdis/cloud-automation.git", branch: 'fix/variable-state'
```
philloooo: the branch should be just whatever branch of active prs that jenkins found?

tf_files/aws/terraform.tfvars
frickjack: If we have that file, then I think terraform automatically includes it, and you need to set those environment variables.  I thought it was better to move the .tfvars file out to external state directory - ex:
     ~/.gen3/Secrets/terraform/planx-pla.net/config.tfvars
, then run terraform referencing those vars
     terraform --var-file ...
, and backup that file to S3 alongside the terraform state - ex in cdistest:
     s3://cdis-terraform-state/planx-pla.net/v1/config.tfvars
, so everyone knows where those vars are defined, and backed up, and
jenkins can grab them, etc

pack_and_terraform_aws.bash
frickjack: The reason we were setting those variales was to plug into the terraform.tfvars file that I erased which has lines like:
      access_key = "${AWS_ACCESS_KEY}"
  Terraform already has a mechanism to set variable via environment variables named TF_VAR_[name]
...

Jenkins/Pipelines/TerraformPlanOnly/Jenkinsfile
frickjack: You're working on VPC migration.
I thought a good strategy to migrate is to do something like this:
* setup a v2/ VPC
* copy state from v1/ VPC
* tear down v1/ VPC

So the v2/ VPC starts with an empty state.

tf_files/aws/terraform.tfvars
philloooo: the file is rendered to [$creds_dir/terraform.tfvars](https://github.com/uc-cdis/cloud-automation/pull/93/commits/1943eb611b69e5dd168c3e3f7b632b33fe07e3a1#diff-40a95469d103a81a11be31908b3fc11eR263) and then used by [terraform init -backend-config=$creds_dir/terraform.tfvars](https://github.com/uc-cdis/cloud-automation/pull/93/commits/1943eb611b69e5dd168c3e3f7b632b33fe07e3a1#diff-40a95469d103a81a11be31908b3fc11eR264), and you can just backup the whole creds_dir to s3, so I think it's doing what you said?

Jenkins/Pipelines/Terraform/Jenkinsfile
frickjack: This pipeline is configured to run by hand.  
Once we merge can switch to 'master' branch.
The idea is we could run terraform through a jenkins job rather than by hand, but that may not make sense.
The TerraformPlanOnly/JenkinsFile is configured to run with a 'MultiBranch' setup - it has a
    scm checkout
line ...

pack_and_terraform_aws.bash
philloooo: yeah I know, you are exporting the **s3** access_key to the TF_VAR_access_key which can be used by both the s3 backend and deployment. but we have both AWS_ACCESS_KEY and AWS_S3_ACCESS_KEY so I I think it's cleaner to use the var files and specify which var file to use instead of tf variables

tf_files/aws/variables.tf
frickjack: ok

Jenkins/Pipelines/TerraformPlanOnly/Jenkinsfile
philloooo: if we want to use a version number, we can use a git tag?

Jenkins/Pipelines/Terraform/Jenkinsfile
philloooo: I see, so this is the first clone that will only be run once when we setup jenkins?
let's switch to master before we merge?

tf_files/aws/terraform.tfvars
frickjack: no - its a static file - the path to the file is tf_files/aws/terraform.tfvars
the pack_and_terraform.sh script sets a bunch of environment variables, then:

```
cd tf_files/aws
    ../../terraform init
    ../../terraform plan -var-file=$creds_dir/tf_variables -state=$creds_dir/terraform.tfstate
    ../../terraform apply -var-file=$creds_dir/tf_variables -state=$creds_dir/terraform.tfstate
```

frickjack: you're thinking of creds_dir/tf_variables:


        INDEXD=$INDEXD \
        envsubst < tf_files/aws/variables.template >$creds_dir/tf_variables

frickjack: we should wire things up to have something like 'creds_dir/' for each environment + helper scripts to automate saving/fetching .tfvars from S3

philloooo: it's not the same of the tf_variables, the file is render to `~/.creds/{vpc}/terraform.tfvars` so that static template isn't used for `terraform init`, it's confusing because there was a [conflict resolve](https://github.com/uc-cdis/cloud-automation/pull/81/files#diff-38bf8a156d67547e100e73d948940e58L204) that removed rendering the file and also specifying the .tfvar file in the namespaced creds dir...

Jenkins/Pipelines/TerraformPlanOnly/Jenkinsfile
frickjack: yeah - that's cool.  That's just what I was thinking when I set things up to save state there.  What would the git tag correspond to?  We may not want to migrate to a new VPC for every change to our terraform config - some things (like add elasitc search) we can just do in place ...

Jenkins/Pipelines/Terraform/Jenkinsfile
frickjack: we point Jenkins at which Jenkinsfile we want it to run when we configure a 'job', and there are different kinds of jobs with different rules.  Let's do a walk-through of what I have setup next week.  It's just v1 - definitely need to iterate :-)

Jenkins/Pipelines/TerraformPlanOnly/Jenkinsfile
philloooo: would be corresponding to each release (which we haven't done any yet). Yeah we don't need to migrate new vpc for every change, I guess when a pr is merged, the changes can be applied to the test vpc `v1.1` and when we create a new git tag jenkins can do a actual vpc swap

Jenkins/Pipelines/Terraform/Jenkinsfile
philloooo: yep!

PR #112, opened 2017-12-05
feat(fence): change from user-api to fence

rudyardrichter:


tf_files/aws/kube.tf
```diff
@@ -35,18 +35,18 @@ resource "aws_subnet" "public_kube" {
 }
 
 
-resource "aws_db_instance" "db_userapi" {
+resource "aws_db_instance" "db_fence" {
```
frickjack: We should try to make our cloud-formation changes backward compatible, so we can easily apply them to existing commons, and deploy fence without undeploying user-api, so we can have an easy rollback.
Rather than remove the 'db_userapi' RDS instance -leave that in place, and add a new 'db_fence' instance, so we can run the 2 side by side for a while.  At the point that a commons is completely transitioned over to fence, then we can remove the userapi db later by setting a 'userapi_count' varialbe that defaults to 1 to 0:
     https://blog.gruntwork.io/terraform-tips-tricks-loops-if-statements-and-gotchas-f739bbae55f9

pack_and_terraform_aws.bash
```diff
@@ -278,11 +278,11 @@ if echo "$runtf" | grep -iq "^y"; then
         google_client_id=$google_client_id \
         hmac_encryption_key=$hmac_encryption_key \
         gdcapi_secret_key=$gdcapi_secret_key \
-        db_password_userapi=$db_password_userapi \
+        db_password_fence=$db_password_fence \
         db_password_indexd=$db_password_indexd \
         db_password_gdcapi=$db_password_gdcapi \
         gdcapi_indexd_password=$gdcapi_indexd_password \
-        userapi_snapshot=$userapi_snapshot \
```
frickjack: See the comment on the aws_db_instance in kube.tf - do not remove userapi - just add fence, so we can deploy fence to existing commons without blowing up the service that are running in place, and we can easily roll back if problems come up.

tf_files/aws/variables.tf
```diff
@@ -24,13 +24,13 @@ variable "base_ami" {
...
 }
-variable "db_password_userapi" {
+variable "db_password_fence" {
 }
 variable "db_password_gdcapi" {
 }
 variable "db_password_indexd" {
 }
-variable "userapi_snapshot" {
+variable "fence_snapshot" {
```
frickjack: Give the fence_snapshot a default value of "" - the snapshot only comes into play when we migrate the database from an old VPC to a new VPC.

pack_and_terraform_aws.bash
```diff
@@ -188,18 +188,6 @@ if echo "$runtf" | grep -iq "^y"; then
         [ -z "$aws_s3_bucket" ] && aws_s3_bucket="cdis-terraform-states"
     fi
 
-    if [ -z "$base_ami" ]; then
```
philloooo: why are these removed?

tf_files/configs/kube-setup-fence.sh
```diff
@@ -0,0 +1,89 @@
...
+  kubectl create secret generic fence-secret --from-file=local_settings.py=./apis_configs/fence_settings.py
+fi
+
+if ! kubectl get secrets/fence-jwt-keys > /dev/null 2>&1; then
+  kubectl create secret generic fence-jwt-keys --from-file=./jwt-keys
+fi
+
+kubectl apply -f services/fence/fence-deploy.yaml
+
+if [[ -z "${fence_snapshot}" && "${create_fence_db}" = "true" && ( ! -f .rendered_fence_db ) ]]; then
```
philloooo: I didn't see `create_fence_db`templated by terraform?

pack_and_terraform_aws.bash
frickjack: terraform has switched over to copying the public ubuntu16 images - those variables are no longer present in the .tf files

tf_files/configs/kube-setup-fence.sh
frickjack: it's hard-coded to 'true' in kube-services.sh (via kube-services-body.sh):
    create_fence_db="true"

The assumption is that you'll run 'kube-services.sh' to setup a brand new VPC
otherwise - if you're running kube-setup-fence.sh directly, then the assumption
is that you're running kube-setup-fence.sh directly, so you're probably updating
and existing environment that already has a userapi db that fence will use.

You can always do:
   create_fence_db=true bash kube-setup-fence.sh vpcname

Note sure how to link to the changes: kube.tf :

     provisioner "local-exec" {
-        command = "echo \"${data.template_file.kube_vars.rendered}\" | cat - \"${path.module}/../configs/kube-certs.sh\" \"${path.module}/../configs/kube-services-body.sh\" > ${var.vpc_name}_output/kube-services.sh"
+        command = "echo \"${data.template_file.kube_vars.rendered}\" | cat - \"${path.module}/../configs/kube-setup-certs.sh\" \"${path.module}/../configs/kube-services-body.sh\" \"${path.module}/../configs/kube-setup-fence.sh\" > ${var.vpc_name}_output/kube-services.sh"
     }

kube-services-body.sh:

+#
+# Set a flag for the kube-setup-fence fragment
+# that terraform appends to kube-services.sh
+#
+create_fence_db="true"
+

PR #160, opened 2018-02-09
Chore/migrate fixes

frickjack:
some post bhcv2 fixes and tweaks ...

* small bugs that popped up deploying bhcprodv2
    - SQL typo
    - deploy reverse proxy last
   - allow major postgres upgrade from snapshot
* remove quay secrets - docker registries are now public
* terraform adds kube-ssh-key to ec2 VM's

### DIFF
--- a/tf_files/configs/kube-setup-sheepdog.sh
+++ b/tf_files/configs/kube-setup-sheepdog.sh
@@ -65,7 +65,7 @@ for user in sheepdog peregrine; do
       psql -t -U $gdcapi_db_user -h $gdcapi_db_host -d $gdcapi_db_database -c "CREATE USER $new_db_user WITH PASSWORD '$new_db_password';"
       psql -t -U $gdcapi_db_user -h $gdcapi_db_host -d $gdcapi_db_database -c "GRANT SELECT ON ALL TABLES IN SCHEMA public TO $new_db_user;"
       if [[ "$new_db_user" =~ ^sheepdog ]]; then
-        psql -t -U $gdcapi_db_user -h $gdcapi_db_host -d $gdcapi_db_database -c "GRANT ALL ON ALL TABLES IN SCHEMA pubic TO $new_db_user;"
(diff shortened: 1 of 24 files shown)


tf_files/configs/kube-setup-sheepdog.sh
```diff
@@ -65,7 +65,7 @@ for user in sheepdog peregrine; do
       psql -t -U $gdcapi_db_user -h $gdcapi_db_host -d $gdcapi_db_database -c "CREATE USER $new_db_user WITH PASSWORD '$new_db_password';"
       psql -t -U $gdcapi_db_user -h $gdcapi_db_host -d $gdcapi_db_database -c "GRANT SELECT ON ALL TABLES IN SCHEMA public TO $new_db_user;"
       if [[ "$new_db_user" =~ ^sheepdog ]]; then
-        psql -t -U $gdcapi_db_user -h $gdcapi_db_host -d $gdcapi_db_database -c "GRANT ALL ON ALL TABLES IN SCHEMA pubic TO $new_db_user;"
+        psql -t -U $gdcapi_db_user -h $gdcapi_db_host -d $gdcapi_db_database -c "GRANT ALL ON ALL TABLES IN SCHEMA public TO $new_db_user;"
```
jbarno: 🙏 we didn't have that SCHEMA before.

tf_files/configs/kube-setup-sheepdog.sh
```diff
@@ -65,7 +65,7 @@ for user in sheepdog peregrine; do
       psql -t -U $gdcapi_db_user -h $gdcapi_db_host -d $gdcapi_db_database -c "CREATE USER $new_db_user WITH PASSWORD '$new_db_password';"
       psql -t -U $gdcapi_db_user -h $gdcapi_db_host -d $gdcapi_db_database -c "GRANT SELECT ON ALL TABLES IN SCHEMA public TO $new_db_user;"
```
philloooo: we should create a group user and add those users to a group, and then use [graph-grant](https://github.com/uc-cdis/datamodelutils/blob/master/datamodelutils/postgres_admin.py#L483) to grant the group user access. Also need to add the graph-grant to our database migration protocol ( didn't have that because the creator was the only user)

PR #166, opened 2018-02-13
Chore/logging

### LINKED ISSUE #152: log user id in access log
would be great to be able to decode jwt token and get user id and inject to access log

frickjack:
* log client IP address (not ELB), and forward to proxied services
* log user (from JWT token)

resolve #152

### DIFF
--- a/kube/services/revproxy/00nginx-config.yaml
+++ b/kube/services/revproxy/00nginx-config.yaml
@@ -39,8 +39,11 @@ data:
       
       	##
       	# Logging Settings
-      	##      
-      	access_log /dev/stdout;
+      	##
+        log_format aws   '$http_x_forwarded_for - $remote_user [$time_local] '
+                    '"$request" $status $body_bytes_sent '
+                    '"$http_referer" "$http_user_agent"';
+      	access_log /dev/stdout aws;
       	error_log /dev/stderr;
       
       	##
--- a/kube/services/revproxy/00nginx-config.yaml
+++ b/kube/services/revproxy/00nginx-config.yaml
@@ -8,12 +8,72 @@ metadata:
   name: revproxy-nginx-conf
   namespace: default 
 data:
+  userid.js: |
+      /** global supporting atob polyfill below */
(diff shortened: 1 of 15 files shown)


kube/services/revproxy/00nginx-config.yaml
```diff
@@ -8,12 +8,72 @@ metadata:
...
+        var user = "unknown@unknown";
+
+        if (token) {
+          user = token;
+          var raw = atob((token.split('.')[1] || "").replace('-', '+').replace('_', '/'));
+          if (raw) {
+            try {
+              var data = JSON.parse(raw);
+              if (data && data.context && data.context.user && data.context.user.name) {
+                user = data.context.user.name;
```
philloooo: can we log both username and user id? (`data.sub`)

kube/services/revproxy/00nginx-config.yaml
```diff
@@ -80,6 +150,8 @@ data:
           #   not set headers ... http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_set_header
           #
           proxy_set_header   Authorization "$access_token";
+          proxy_set_header   X-Forwarded-For "$http_x_forwarded_for";
+          proxy_set_header   X-UserId "$userid";
           #add_header X-debug-auth "$access_token"; #- only do this for quick and dirty client-side debugging
```
philloooo: can remove this now?

kube/services/revproxy/00nginx-config.yaml
```diff
@@ -102,47 +174,43 @@ data:
           }
           
           location / {
-              set $portal_service http://portal-service.default.svc.cluster.local;
-              if ($cookie_portal_service ~ ^[\w-]+$) {
```
philloooo: why is this removed?

frickjack: I want to replace with changes so we can deploy the whole commons into a namespace rather than just do the portal-switch with a cookie thing - which I don't think has turned out to be too useful ...?

kube/services/revproxy/00nginx-config.yaml
frickjack: Need to pass the IP address data and user data on to the child services, so they can log it ...

PR #212, opened 2018-03-01
feat(jupyterhub): JupyterHub Service

### LINKED ISSUE #173: Test persistent disk w/ /home/jovyan for jupyter notebooks so we don't need /home/jovyan/pd

### LINKED ISSUE #174: Define default docker image for Jupyter with tools we want

### LINKED ISSUE #175: Define default resource constraints for user pods

### LINKED ISSUE #176: Create separate namespace for user pods

zflamig:
Adds the configs for the jupyterhub service
Adds the shell script to install jupyterhub service
Not automatically installed on new commons
Resolves #176
Resolves #175 
Resolves #174 
Resolves #173 

To actually enable jupyterhub we also need to switch the ELB from HTTPS to TCP right now.

### DIFF
--- a/kube/services/jupyterhub/jupyterhub-config.yaml
+++ b/kube/services/jupyterhub/jupyterhub-config.yaml
@@ -38,7 +38,7 @@ data:
             'name': 'volume-{username}{servername}'
         }
     ]
-    c.KubeSpawner.hub_connect_ip = 'jupyterhub-service.default'
+    c.KubeSpawner.hub_connect_ip = 'jupyterhub-service.%s' % (os.environ['POD_NAMESPACE'])
     c.KubeSpawner.hub_connect_port = 8000
     c.KubeSpawner.singleuser_image_spec = 'quay.io/occ_data/jupyternotebook:latest'
     c.KubeSpawner.cmd = 'start-singleuser.sh'
--- a/kube/services/jupyterhub/jupyterhub-config.yaml
+++ b/kube/services/jupyterhub/jupyterhub-config.yaml
@@ -13,7 +13,7 @@ data:
     #c.ConfigurableHTTPProxy.debug = True
     c.JupyterHub.log_level = 'DEBUG'
     c.JupyterHub.spawner_class = 'kubespawner.KubeSpawner'
-    c.KubeSpawner.namespace = 'jupyter_pods'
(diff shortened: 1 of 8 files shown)


kube/services/jupyterhub/jupyterhub-deployment.yaml
```diff
@@ -0,0 +1,55 @@
+apiVersion: apps/v1beta1
+kind: StatefulSet
+metadata:
+  name: jupyterhub-deployment
+spec:
+  replicas: 1
```
frickjack: Remove replicas.
Default is 1, and we can scale up via `patch_kreps` function in kubes.sh, and not worry about re-applying a deployment.yaml

kube/services/revproxy/00nginx-config.yaml
```diff
@@ -236,5 +242,35 @@ data:
               send_timeout 300;
               proxy_pass http://sheepdog-service/;
           }
+          location /workspace/ {
```
frickjack: I thought you needed to localize the csrf check?

kube/services/revproxy/00nginx-config.yaml
```diff
@@ -236,5 +242,35 @@ data:
...
+              auth_request /auth-proxy;
+              auth_request_set $remoteUser $upstream_http_REMOTE_USER;
+              proxy_set_header REMOTE_USER $remoteUser;
```
frickjack: What is $upstream_http_REMOTE_USER - could you put some comments here explaining what's going on?

kube/services/jupyterhub/jupyterhub-config.yaml
```diff
@@ -0,0 +1,54 @@
...
+            }
+        }
+    ]
+    c.KubeSpawner.volume_mounts = [
+        {
+            'mountPath': '/home/jovyan/pd',
+            'name': 'volume-{username}{servername}'
+        }
+    ]
+    c.KubeSpawner.hub_connect_ip = 'jupyterhub-service.default'
```
frickjack: Do we need to hard code the '.default' namespace?
How is this going to work with namespaces for qa, dev, and red-black deployment?
This is actually a 3rd party package that we don't update as part of our core code - let's just move it out to it's own namespace rather than run in 'default' ...

Issue #201, opened 2018-02-27
setup test ftp docker service


philloooo:
setup a test ftp server that has similar user list and same encryption as dbgap, just with dummy users
the ftp server would be a separate k8s pod and service under qa vpc with a separate elb/alb

philloooo:
the real dbgap should only be tested against in Commons specific qa/staging environment for commons that need to use dbgap

thanh-nguyen-dang:
really spent 8 hours

PR #223, opened 2018-03-14
Feat/sftp

thanh-nguyen-dang:
r @frickjack


kube/services/sftp/sftp-config.yaml
```diff
@@ -0,0 +1,21 @@
+#
+# Note: apply this file via apply_config to fill in the DNS
```
zflamig: Is this comment meaningful to us? I think its just been copy+pasted forever..

thanh-nguyen-dang:
resolve #201

PR #232, opened 2018-03-17
chore(jh): Add user pod culling service

zflamig:
chore(revproxy): Setup as ssl for websockets support


kube/services/revproxy/revproxy-service-elb.yaml
```diff
@@ -0,0 +1,20 @@
...
+    service.beta.kubernetes.io/aws-load-balancer-proxy-protocol: "*"
+    # supported in k8s 1.9
+    service.beta.kubernetes.io/aws-load-balancer-ssl-negotiation-policy: "ELBSecurityPolicy-TLS-1-2-2017-01"
+spec:
+  selector:
+    app: revproxy
+  ports:
+    - protocol: TCP
+      port: 443
+      targetPort: 81
```
philloooo: who has port 81 open?

kube/services/revproxy/revproxy-service.yaml
```diff
@@ -2,23 +2,12 @@ kind: Service
...
-    # right now kubernete make them both https for the elb protocol, has to manually fix it after you apply this file
-    - protocol: TCP 
-      port: 80 
-      targetPort: 80 
+    - protocol: TCP
+      port: 80
+      targetPort: 80
       name: http
-  type: LoadBalancer 
+  type: NodePort
```
philloooo: how is this nodeport exposed out?

kube/services/revproxy/revproxy-service-elb.yaml
zflamig: So using SSL TCP on ELB breaks the http forwarded for headed, which means logging can't log the client IP. We fix this by using proxy protocol on the ELB. RevProxy Nginx then exposes port 81 which reads the proxy protocol + services requests. So the ELB -> revproxy nginx happens on port 81. Internal traffic still talks to revproxy on port 80 since it isn't sending the proxy protocol headers.

kube/services/revproxy/apply_service
```diff
@@ -4,7 +4,9 @@ scriptDir=$(dirname -- "$(readlink -f -- "$BASH_SOURCE")")
 
 export ARN=$(kubectl get configmap global --output=jsonpath='{.data.revproxy_arn}')
 if [[ ! -z $ARN ]]; then
-  envsubst <$scriptDir/revproxy-service.yaml | kubectl apply -f -
+  envsubst <$scriptDir/revproxy-service-elb.yaml | kubectl apply -f -
 else
   echo "Global configmap not configured"
 fi
+
+kubectl apply -f $scriptDir/revproxy-service.yaml
```
philloooo: let's move apply the old yaml out so it doesn't remove elb before dns is switched

PR #240, opened 2018-03-21
Feat/adminvm

### LINKED ISSUE #181: setup vpc peering

### LINKED ISSUE #215: setup central vpn

### LINKED ISSUE #238: setup 'adminvm' for CSOC
We have automation to create an 'adminvm' for a commons' AWS account, but we should also create an 'adminvm' for the CSOC itself where we can run `terraform` and other scripts to manage AWS resources in the CSOC account itself.

frickjack:
resolve #238 
resolve #215 
resolve #181


gen3/bin/workon.sh
```diff
@@ -143,13 +143,36 @@ EOM
...
+  env_public_subnet_id = "VPC-PUBLIC-SUBNET"
 EOM
     return 0
   fi
   
+  if [[ "$GEN3_WORKSPACE" =~ _logging$ ]]; then
+    # rds snapshot vpc is simpler ...
+    commonsName=$(echo "$GEN3_WORKSPACE" | sed 's/_logging$//')
+    cat - <<EOM
+child_account_id=""
```
frickjack: child_account_id="NUMERIC_ID_HERE"

gen3/bin/workon.sh
```diff
@@ -143,13 +143,36 @@ EOM
...
+common_name=""
```
frickjack: commons_name="$commonsName"

tf_files/aws_squid_vm/outputs.tf
```diff
@@ -0,0 +1,5 @@
+output "squid_public_ip" {
+ value = "${module.squid_vm.squid_public_ip}"
+}
+
+
```
frickjack: Do you need the 'private_ip' as an output too, so it can be used in internal route tables?

tf_files/modules/cdis-aws-common-logging/lambda_function.py
```diff
@@ -113,17 +110,9 @@ def handler(event, context):
     client = boto3.client('firehose')
```
frickjack: Would you improve the doc-comments on every function and on the file as a while.  In particular - it would be good to know what the format of the input and output to each function is (ex: { "message" : "a comment line", "timestamp": "iso-formated utc timestamp" }), and whether the function modifies the input (like `dateIt` does).

Also - it would be good to know the overall flow - ex 
  - I assume that `handler` is invoked as new message bundles arrive into the Kinesis stream
  - I assume the output of `handler` gets injected into the S3 and Elastic Search fire hoses?

frickjack: The python style guide probably has more tips that I didn't notice :-p :
   https://github.com/uc-cdis/rfcs/blob/master/text/0003-style-guide.md#python-style-guide

tf_files/modules/cdis-aws-common-logging/logging.tf
```diff
@@ -1,12 +1,12 @@
 # We need a bucket so we can upload logs from Elastic Search, logs from the child account, and
 # Kinesis stream logs
 
-resource "aws_s3_bucket" "child_account_bucket" {
-  bucket = "${var.child_name}-logging"
+resource "aws_s3_bucket" "common_logging_bucket" {
+  bucket = "${var.common_name}-logging"
```
frickjack: We need to enforce that bucket names do not include '_' or '.' - take a look at the `local` block in the `cdis-s3-bucket` module:
```
locals {
  clean_bucket_name = "${replace(replace(var.bucket_name, "_", "-"),".", "-")}"
}
```

tf_files/modules/cdis-aws-common-logging/logging.tf
```diff
@@ -189,77 +185,73 @@ data "aws_iam_policy_document" "firehose_policy_document" {
...
   policy = "${data.aws_iam_policy_document.firehose_policy_document.json}"
   role   = "${aws_iam_role.firehose_role.id}"
 }
 
 # Need these guys because the firehose resource is not that smart to create it if it doesn't exist
 
-resource "aws_cloudwatch_log_group" "csoc_log_group" {
-  name = "${var.child_name}"
+resource "aws_cloudwatch_log_group" "csoc_common_log_group" {
+  name = "${var.common_name}"
```
frickjack: So we have a log-group per commons in the CSOC?  What logs go there - just the `adminvm`?  
It looks like the `firehost_to_s3` is associated with this log_group?  Does the firehose drop logs here, then?
Maybe just add a comment explaining what's going on for people like me ... :-p

tf_files/modules/cdis-aws-common-logging/README.md
```diff
@@ -0,0 +1,70 @@
...
+      "awsRegion": "us-east-1"
+    }
+  ]
+}
+```
+[Table 1] Example fo data received through streaming service
+
+The actual data from child accountd is in the `data` part of the above JSON, the rest is added by the service which is also nice to have. The lamda function takes care of decoding the data and make it nice for ElasticSearch.
+
+Logs are being sent to S3 as well. There are multiple ways to store the data into S3, as it comes from the stream, or processed the same way we do for ElasticSearch. As for now, we are processing it so we don't need to decode/uncompress if we want/need to see the content.
```
frickjack: Maybe add a note that each commons VPC gets its own elastic search index.

tf_files/modules/cdis-aws-common-logging/logging.tf
fauzigo: using '-'. I found out the hardway that S3 doesn't like '.' or '_'.

This should be fine now.

tf_files/modules/cdis-aws-common-logging/logging.tf
fauzigo: Commenting the code, but basically firehoses would put anything there if it errors for whatever reason and cant' reach the destination.

Have not happened yet, but that should help understanding if firehoses fail

tf_files/modules/cdis-aws-common-logging/README.md
fauzigo: right on

tf_files/aws_squid_vm/outputs.tf
rarya: I don't need it for the route tables; but probably good to have it as a output for ssh access.I will add it.

tf_files/modules/cdis-aws-common-logging/README.md
fauzigo: done

tf_files/modules/cdis-aws-common-logging/lambda_function.py
fauzigo: more documentation in the lambda_function.py itself

gen3/bin/workon.sh
fauzigo: DONE

gen3/bin/workon.sh
fauzigo: done

PR #247, opened 2018-04-04
Chore/jenkins qa

frickjack:
Implement `kube-role-qa.sh` with supporting changes.

See also: https://github.com/uc-cdis/cdis-manifest/blob/QA/Jenkinsfile

### DIFF
--- a/kube/services/fence/fence-deploy.yaml
+++ b/kube/services/fence/fence-deploy.yaml
@@ -19,6 +19,18 @@ spec:
         app: fence
         GEN3_DATE_LABEL
     spec:
+      affinity:
+        podAntiAffinity:
+          preferredDuringSchedulingIgnoredDuringExecution:
+          - weight: 100
+            podAffinityTerm:
+              labelSelector:
+                matchExpressions:
+                - key: app
+                  operator: In
+                  values:
+                  - fence
+              topologyKey: "kubernetes.io/hostname"
       automountServiceAccountToken: false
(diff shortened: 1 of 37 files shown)


kube/services/fence/fence-deploy.yaml
```diff
@@ -19,6 +19,18 @@ spec:
         app: fence
         GEN3_DATE_LABEL
     spec:
+      affinity:
```
philloooo: this is trying to make pods for particular app don't be scheduled on the same node?

frickjack: yep

tf_files/configs/kube-roll-qa.sh
```diff
@@ -0,0 +1,47 @@
...
+
+result=0
+namespaceList=$(g3kubectl get namespace -o json | jq -r '.items[].metadata.name')
+for name in $namespaceList; do
+  if [[ "$name" == "default" || $name =~ ^qa-.+$ ]]; then
+    echo $name
+    if ! (
+      export WORKSPACE=$(mktemp -d -p "$WORKSPACE" "qaroll_${name}_XXXXXX")
+      # g3kubectl keys on KUBECTL_NAMESPACE environment variable
+      export KUBECTL_NAMESPACE="$name"
```
philloooo: can you comment about this actually determines which manifest to use?

PR #248, opened 2018-04-06
Feat/fence dbgap

### LINKED ISSUE #253: create job for dbgap syncing

giangbui:
resolve #253


tf_files/configs/kube-setup-fence.sh
```diff
@@ -62,12 +68,57 @@ if [[ -d "${WORKSPACE}/${vpc_name}_output" ]]; then # update secrets
...
+
+        Host cloud-proxy.internal.io
+          StrictHostKeyChecking no
+          UserKnownHostsFile=/dev/null
+        ''' > ./apis_configs/.ssh/config
+    fi
+    kubectl create configmap fence-sshconfig --from-file=./apis_configs/.ssh/config
+  fi
+
+  if ! kubectl get secrets/fence-public-key > /dev/null 2>&1; then
```
frickjack: Can we combined fence-public-key and fence-ssh-key into one secret - fence-ssh-keys ?

tf_files/configs/kube-extract-config.sh
```diff
@@ -72,7 +72,14 @@ revproxyArn=$(g3kubectl get configmaps/global -o=jsonpath='{.data.revproxy_arn}'
...
   for keyName in jwt_public_key.pem jwt_private_key.pem; do
-    g3kubectl get secrets/fence-jwt-keys -o=json | jq -r ".data[\"$keyName\"]" | base64 --decode > "./jwt_keys/$keyName"
+    g3kubectl get secrets/fence-jwt-keys -o=json | jq -r ".data[\"$keyName\"]" | base64 --decode > "./jwt-keys/$keyName"
+  done
+fi
+
+if g3kubectl get secrets/fence-private-key > /dev/null 2>&1; then
+  mkdir -p -m 0700 ./ssh-keys
+  for keyName in id_rsa id_rsa.pub; do
+    g3kubectl get secrets/fence-private-key -o=json | jq -r ".data[\"$keyName\"]" | base64 --decode > "./ssh-keys/$keyName"
```
frickjack: Switch to `fence-ssh-keys` - see below ...

tf_files/configs/kube-setup-fence.sh
```diff
@@ -62,12 +68,57 @@ if [[ -d "${WORKSPACE}/${vpc_name}_output" ]]; then # update secrets
...
-  if ! g3kubectl get secrets/fence-jwt-keys > /dev/null 2>&1; then
-    g3kubectl create secret generic fence-jwt-keys --from-file=./jwt-keys
+  if ! kubectl get configmaps/projects > /dev/null 2>&1; then
+    if [[ ! -f "./apis_configs/projects.yaml" ]]; then
+      touch "apis_configs/projects.yaml"
+    fi
+    kubectl create configmap projects --from-file=apis_configs/projects.yaml
+  fi
+
+  if ! kubectl get configmaps/user-dir > /dev/null 2>&1; then
```
frickjack: What is `user-dir` ?

tf_files/configs/kube-setup-fence.sh
```diff
@@ -62,12 +68,57 @@ if [[ -d "${WORKSPACE}/${vpc_name}_output" ]]; then # update secrets
...
+
+        Host sftp.planx
+          ServerAliveInterval 120
+          HostName sftp.planx-pla.net
+          User foo
+          ForwardAgent yes
+          IdentityFile ~/.ssh/id_rsa
+          ProxyCommand ssh ubuntu@squid.internal nc %h %p 2> /dev/null
+
+        Host cloud-proxy.internal.io
```
frickjack: Should we have an entry for sftp.dbgap ?

tf_files/configs/kube-setup-fence.sh
giangbui: Fixed at 949c772f16eef81339337b42dbf68fb52f6ae903
Thanks

tf_files/configs/kube-extract-config.sh
giangbui: Fixed at 949c772

tf_files/configs/kube-setup-fence.sh
giangbui: In production deployment, we will replace it with sftp.dbgap. Right?

frickjack: We don't want to have to add the dbgap entry by hand.  Is there any reason not to put an entry here?

giangbui: We setup sftp.planx-plan.net so that we don't have to manipulate with real sftp dbgap. I don't think we use sftp.planx for other purpose.
We can edit the file ./apis_configs/.ssh/config to configure with the new sftp server
Anyway we can add that entry there

kube/services/fence/fence-deploy.yaml
```diff
@@ -68,6 +80,14 @@ spec:
...
         - containerPort: 80
         - containerPort: 443
+        lifecycle:
+          postStart:
+            exec:
+              command:
+                - bash
+                - -c
+                - cp /fence/ssh-keys/..data/* /root/.ssh/ && chmod 400 /root/.ssh/id_rsa /root/.ssh/id_rsa.pub
+
```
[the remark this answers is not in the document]
giangbui: Yeap. Since we want to copy the physical files not links. I tried with copying links but chmod command does not work on links

kube/services/jobs/dbgapsync-job.yaml
```diff
@@ -0,0 +1,65 @@
+apiVersion: batch/v1
```
philloooo: I don't think we should create a new sync job? the existing usersync job and usersync cronjob should be changed to use the fence-create script, and use the current user yaml either from the local yaml or s3, and optionally sync from dbgap depends on a flag in configmap?

Issue #249, opened 2018-04-06
Network:  Automate the configuration of VPC network peering for new VPC/Projects


rpowellgit:

rarya:
I wrote a bash script `approve_vpcpeering_request.sh` which can be run on the csoc admin VM. It looks for any pending request for VPC peering, if any it accepts the request, name tags the VPC peering connection and adds the appropriate route on the private subnet routing table of the csoc main vpc.

PR #254, opened 2018-04-09
chore(vpcpeering): automating VPC peering request at csoc admin VM

rarya:
Adds a bash script [approve_vpcpeering_request.sh] to accept the peering request and add the appropriate route.


rarya:
#249

PR #255, opened 2018-04-09
Feat/google configuration

Avantol13:
- Parse settings for google configuration
- setup a google credentials secret to access project and group management
    - this houses the google creds for a service account
- k8s cronjobs for scheduled monitoring and account expiration
- k8s jobs for monitoring account expiration
    - separate jobs (in addition to cronjobs) makes it easier to run by itself outside of the scheduled time and also makes testing easier

Depends on https://github.com/uc-cdis/fence/pull/176


apis_configs/fence_settings.py
```diff
@@ -68,6 +68,17 @@
         ENABLED_IDENTITY_PROVIDERS = data['ENABLED_IDENTITY_PROVIDERS']
         APP_NAME = data['APP_NAME']
         HTTP_PROXY = data['HTTP_PROXY']
+        os.environ["GOOGLE_API_KEY"] = data['GOOGLE_API_KEY']
```
Avantol13: Should this be a k8s secret as well?

frickjack: `fence_credentials.json` is already a secret, so this should be good ...

apis_configs/fence_settings.py
```diff
@@ -68,6 +68,17 @@
...
+        os.environ["GOOGLE_IDENTITY_DOMAIN"] = data['GOOGLE_IDENTITY_DOMAIN']
+        os.environ["GOOGLE_CLOUD_IDENTITY_ADMIN_EMAIL"] = (
+            data['GOOGLE_CLOUD_IDENTITY_ADMIN_EMAIL']
+        )
+
+os.environ["GOOGLE_APPLICATION_CREDENTIALS"] = (
+    "/var/www/fence/google_secret.json"
+)
 
 DEFAULT_LOGIN_URL_REDIRECT_PARAM = 'redirect'
```
frickjack: Should we replace `data[X]` with `data.get(X)`, so it doesn't explode if the json configuration is missing some keys.  Even better - fall back to reasonable default values, and maybe log a message ...?

tf_files/configs/kube-setup-cronjobs.sh
```diff
@@ -0,0 +1,19 @@
...
+
+set -e
+
+_KUBE_SETUP_CRONJOBS=$(dirname "${BASH_SOURCE:-$0}")  # $0 supports zsh
+# Jenkins friendly
+export WORKSPACE="${WORKSPACE:-$HOME}"
+export GEN3_HOME="${GEN3_HOME:-$(cd "${_KUBE_SETUP_CRONJOBS}/../.." && pwd)}"
+
+if ! kubectl get cronjob google-manage-keys > /dev/null 2>&1; then
+   kubectl create -f "${GEN3_HOME}/kube/services/jobs/google-manage-keys-cronjob.yaml"
```
frickjack: Please rebase on master, and switch to using `g3kubectl` throughout.  The latest code uses `g3kubectl` - which is a wrapper that switches namespaces based on the KUBECTL_NAMESPACE environment.  That setup makes it easier for us to run these scripts in a Jenkins pipeline ...

apis_configs/fence_settings.py
Avantol13: yeah, that's probably a better idea

tf_files/configs/kube-setup-cronjobs.sh
Avantol13: gotcha, will do

apis_configs/fence_settings.py
Avantol13: fixed in 8347fc3

tf_files/configs/kube-setup-fence.sh
```diff
@@ -62,12 +63,28 @@ if [[ -f "${WORKSPACE}/${vpc_name}_output/creds.json" ]]; then # update secrets
...
   if ! g3kubectl get secrets/fence-json-secret > /dev/null 2>&1; then
     if [[ ! -f "./apis_configs/fence_credentials.json" ]]; then
-      cp "${GEN3_HOME}/tf_files/configs/fence_credentials.json" "./apis_configs/fence_credentials.json" 
+      cp "${GEN3_HOME}/tf_files/configs/fence_credentials.json" "./apis_configs/fence_credentials.json"
     fi
     echo "create fence-json-secret using current creds file apis_configs/fence_credentials.json"
     g3kubectl create secret generic fence-json-secret --from-file=fence_credentials.json=./apis_configs/fence_credentials.json
   fi
 
+  if ! kubectl get secrets/fence-google-app-creds-secret > /dev/null 2>&1; then
```
frickjack: Would you switch this to use `g3kubectl` ?  It's a little wrapper that respects the `KUBECTL_NAMESPACE` environment variable.  It's not really necessary here (since Jenkins doesn't create secrets), but I'd like to stay consistent (I have another branch that fixes the other calls :-p)

kube/services/jobs/google-manage-keys-job.yaml
```diff
@@ -0,0 +1,78 @@
...
+            mountPath: "/fence/keys"
+          - name: "config-helper"
+            readOnly: true
+            mountPath: "/var/www/fence/config_helper.py"
+            subPath: config_helper.py
+        command: ["/bin/bash"]
+        args:
+          - "-c"
+          - |
+            fence-create google-manage-keys
```
frickjack: Add another block here:
```
if [[ $? != 0 ]]; then
  echo "WARNING: non zero exit code: $?"
fi
```
So the script always succeeds (always has exit code 0) - just logs failures, and we avoid Kubernetes annoying retry logic kicking in ...

kube/services/jobs/google-manage-keys-cronjob.yaml
```diff
@@ -0,0 +1,85 @@
...
+                mountPath: "/fence/keys"
+              - name: "config-helper"
+                readOnly: true
+                mountPath: "/var/www/fence/config_helper.py"
+                subPath: config_helper.py
+            command: ["/bin/bash"]
+            args:
+              - "-c"
+              - |
+                fence-create google-manage-keys
```
frickjack: Add another block here:
```
if [[ $? != 0 ]]; then
  echo "WARNING: non zero exit code: $?"
fi
```
So the script always succeeds (always has exit code 0) - just logs failures, and we avoid Kubernetes annoying retry logic kicking in ...

kube/services/jobs/google-manage-account-access-job.yaml
```diff
@@ -0,0 +1,78 @@
...
+            mountPath: "/fence/keys"
+          - name: "config-helper"
+            readOnly: true
+            mountPath: "/var/www/fence/config_helper.py"
+            subPath: config_helper.py
+        command: ["/bin/bash"]
+        args:
+          - "-c"
+          - |
+            fence-create google-manage-account-access
```
frickjack: Add another block here:
```
if [[ $? != 0 ]]; then
  echo "WARNING: non zero exit code: $?"
fi
```
So the script always succeeds (always has exit code 0) - just logs failures, and we avoid Kubernetes annoying retry logic kicking in ...

kube/services/jobs/google-manage-account-access-cronjob.yaml
```diff
@@ -0,0 +1,85 @@
...
+                mountPath: "/fence/keys"
+              - name: "config-helper"
+                readOnly: true
+                mountPath: "/var/www/fence/config_helper.py"
+                subPath: config_helper.py
+            command: ["/bin/bash"]
+            args:
+              - "-c"
+              - |
+                fence-create google-manage-account-access
```
frickjack: Add another block here:
```
if [[ $? != 0 ]]; then
  echo "WARNING: non zero exit code: $?"
fi
```
So the script always succeeds (always has exit code 0) - just logs failures, and we avoid Kubernetes annoying retry logic kicking in ...

kube/services/jobs/google-create-bucket-job.yaml
```diff
@@ -0,0 +1,156 @@
...
+                create_command+=" --google-project-id $GOOGLE_PROJECT_ID"
+            fi
+            if [[ -n "$PROJECT_AUTH_ID" ]]; then
+                create_command+=" --project-auth-id $PROJECT_AUTH_ID"
+            fi
+            if [[ -n "$ACCESS_LOGS_BUCKET" ]]; then
+                create_command+=" --access-logs-bucket $ACCESS_LOGS_BUCKET"
+            fi
+
+            fence-create ${create_command}
```
frickjack: Add another block here:
```
if [[ $? != 0 ]]; then
  echo "WARNING: non zero exit code: $?"
fi
```
So the script always succeeds (always has exit code 0) - just logs failures, and we avoid Kubernetes annoying retry logic kicking in ...

Issue #117, opened 2017-12-15
upgrade to kube-aws v0.9.9


frickjack:
looks like kube-aws v0.9.9 has gone GA
that's kubernetes 1.8 - with real RBAC
* https://github.com/kubernetes-incubator/kube-aws/releases/tag/v0.9.9
* https://github.com/kubernetes-incubator/kube-aws/releases

we're running v0.9.8
at least that's what the latest kube-up.sh will try to install

https://github.com/uc-cdis/cloud-automation/blob/master/tf_files/configs/kube-up-body.sh

fauzigo:
v0.9.9 uses 1.8, will be testing and designing an upgrade

fauzigo:
The update is in place, however, we went all the way up to version v0.9.10-rc3. v0.9.9 is still buggy.

We may want to keep this open (or on hold) until they release v0.9.10 GA to apply the update again, which would just be changing the version in the kube-up-body.sh script.

Issue #261, opened 2018-04-12
Change the user variable from vpc_octet to vpc_cidr for launching new data commons


rarya:
As of now, we pass the user variable as vpc_octet which is the third octet for 172.24.Y.0/20 CIDR space. We have total of 16 CIDRs available with this. We want to change this with passing a complete CIDR as 172.X.Y.0/20 instead. This is required as we have almost saturated using 172.24.Y.0/20.

rarya:
This was completed  on 17 Apr 18.

PR #265, opened 2018-04-17
Changed supplying 3rd vpc cidr octet to supplying both 2nd and 3rd octet

rarya:
#261 
This will take care of the saturation for the network space 172.24.X.0/20  for VPC data commons


fauzigo:
Fixes #117

PR #266, opened 2018-04-17
Feat/kube aws update

fauzigo:


fauzigo:
This update assumes that kube-aws is installed already (currently using v0.9.10-rc3, the GA version should go out soon.

Previous versions have a bug that won't allow the cluster deployment. Therefor this is a must update.

tf_files/configs/kube-setup-workvm.sh
```diff
@@ -86,25 +89,25 @@ fi
 EOF
   fi
 
-  if ! grep 'kubectl completion bash' ${WORKSPACE}/.bashrc > /dev/null; then 
-    cat - >>${WORKSPACE}/.bashrc << EOF
+  if ! grep 'kubectl completion bash' ${WORKSPACE}/.${RC_FILE} > /dev/null; then
```
frickjack: This should probably be something like `kubectl completion ${SHELL}`

tf_files/configs/kube-setup-workvm.sh
```diff
@@ -86,25 +89,25 @@ fi
...
   fi
 
-  if ! grep 'kubectl completion bash' ${WORKSPACE}/.bashrc > /dev/null; then 
-    cat - >>${WORKSPACE}/.bashrc << EOF
+  if ! grep 'kubectl completion bash' ${WORKSPACE}/.${RC_FILE} > /dev/null; then 
+    cat - >>${WORKSPACE}/.${RC_FILE} << EOF
 if which kubectl > /dev/null 2>&1; then
   # Load the kubectl completion code for bash into the current shell
-  source <(kubectl completion bash)
+  source <(kubectl completion ${SHELLL})
```
frickjack: SHELLLLLLL typo

tf_files/modules/cdis-aws-squid/variables.tf
```diff
@@ -30,6 +30,14 @@ variable "env_vpc_id" {
...
 }
 
+variable "env_instance_profile" {
+  #default = "common_name_cloudwatch_access_profile"
+}
+
+variable "env_log_group" {
+  #default = "common_name"
+}
+
```
frickjack: Need to update `tf_files/aws_squid_vm/` to account for these changes

tf_files/configs/cluster.yaml
```diff
@@ -42,49 +64,91 @@ worker:
...
+  clusterAutoscaler:
+    enabled: true
+  rescheduler:
+    enabled: false
+  metricsServer:
+    enabled: false
+useCalico: true
+kubelet:
+  RotateCerts:
+    enabled: true
```
frickjack: RotateCerts should be `false` - only applies when `tlsBootstrap` below is `true`, but I think `tlsBootstrap` is flaky ...

tf_files/configs/kube-setup-workvm.sh
fauzigo: SHELLL is not a typo, Linux already uses SHELL to identify the current shell, didn't want to override that env var. I can rename it to something else and less prone to confusion.

tf_files/configs/kube-setup-workvm.sh
fauzigo: good catch, checking

tf_files/configs/kube-setup-workvm.sh
fauzigo: CURRENT_SHELL is used now, looks better.

tf_files/configs/cluster.yaml
fauzigo: changed

Issue #179, opened 2018-02-19
Test NetworkPolicy, get working with Calico?


philloooo:
Use NetworkPolicy in k8s to limit network access for different pods so that user's jupyter pods can't talk to gen3 services except for the revproxy

rpowellgit:
Renuka is going to test her policies in GDC/PDC since its working with calico already. Then we will have them once the AWS k8s clusters are upgraded?

rpowellgit:
Related to #177

rarya:
We have a kubernetes cluster setup in the `raryav1` VPC in cdistest account with updated version of kube-aws.I have done some basic testing to verify that network policy works in our new Calico set-up. I am trying to understand what all services are there and how we need to communicated between them. Created a diagram for the services communication requirement from the network policies.Working on the network policies.

Issue #177, opened 2018-02-19
Write NetworkPolicy rules 


trevars:
for default deny all, revproxy -> services, services -> services, all -> revproxy

PR #297, opened 2018-05-03
network policy for the kubernetes cluster

rarya:
This is the first draft of network policy for the K8s cluster core services
#177 
#179


kube/services/netpolicy/networkpolicy_revproxy_templ.yaml
```diff
@@ -0,0 +1,51 @@
...
+          matchLabels:
+             app: sheepdog
+      - podSelector:
+          matchLabels:
+             app: portal
+      - podSelector:
+          matchLabels:
+             app: fence
+      - ipBlock:
+           cidr: cloudproxy_ip/32
```
frickjack: Would you change this to something like `GEN3_CLOUDPROXY_IP` - so it's more clear that it gets filled in by our template processing, and isn't some kind of kubernetes variable?  Maybe add a comment?  Do the same to the other files?

tf_files/configs/kube-setup-networkpolicy.sh
```diff
@@ -0,0 +1,41 @@
...
+if [[ -z "$_KUBES_SH" ]]; then
+  source "$GEN3_HOME/kube/kubes.sh"
+fi # else already sourced this file ...
+
+vpc_name=${vpc_name:-$1}
+if [ -z "${vpc_name}" ]; then
+   echo "Usage: bash kube-setup-networkpolicy.sh vpc_name"
+   exit 1
+fi
+
```
frickjack: Hi Renuka!

Looks great - let's add a check for the kubernetes version:

```
serverVersion=$(kubectl version server -o json | jq -r '.serverVersion.major + "." + .serverVersion.minor + ".0"')
if ! semver_ge "$serverVersion" "1.8.0"; then
  echo "K8s server version $serverVersion does not yet support network policy"
  exit 0
fi
```

rarya: added with a slight change:
serverVersion="$(kubectl version server -o json | jq -r '.serverVersion.major + "." + .serverVersion.minor' | head -c3).0"
echo "K8s server version is $serverVersion"
if ! semver_ge "$serverVersion" "1.8.0"; then
  echo "K8s server version $serverVersion does not yet support network policy"
  exit 0
fi

PR #308, opened 2018-05-14
feat(webhook): Add slack message to cronjob

zflamig:
Needs `slack_webhook` in the global configmap now.


zflamig:
Hmm, Is there a case for needing the second job still? You can run the cronjobs on demand like `kubectl create job --from=cronjob/usersync slack-test`

PR #313, opened 2018-05-15
Feat/csoc utility vm

fauzigo:


tf_files/aws/modules/utility-vm/cloud.tf
```diff
@@ -0,0 +1,198 @@
...
+
+  egress {
+    from_port   = 0
+    to_port     = 0
+    protocol    = "-1"
+    cidr_blocks = ["10.128.0.0/20", "52.0.0.0/8", "54.0.0.0/8", "${var.vpc_cidr_list}"]
+  }
+
+  tags {
+    Environment = "CSOC"
```
frickjack: Should this be ${var.environment} ?

tf_files/aws/modules/utility-vm/cloud.tf
```diff
@@ -0,0 +1,198 @@
...
+  filter {
+    name   = "name"
+    values = ["ubuntu16-docker-base-1.0.2-*"]
+  }
+
+  owners = ["${var.ami_account_id}"]
+}
+
+resource "aws_ami_copy" "cdis_ami" {
+  name              = "ub16-cdis-crypt-1.0.2-nginx"
```
frickjack: Where does this AMI come from?
Could we just clone a public ubuntu16 image, and `apt install nginx`?
   https://cloud-images.ubuntu.com/locator/ec2/
Or (even better) run a docker? :-)
    https://hub.docker.com/_/nginx/
    https://docs.docker.com/install/linux/docker-ce/ubuntu/#set-up-the-repository

flavors/nginx/es_revproxy.sh
```diff
@@ -0,0 +1,25 @@
...
+datetime_format = %b %d %H:%M:%S
+file = /var/log/syslog
+log_stream_name = syslog-{hostname}-{instance_id}
+time_zone = LOCAL
+log_group_name = ${HOSTNAME}
+EOM
+
+chmod 755 /etc/init.d/awslogs
+systemctl enable awslogs
+systemctl restart awslogs
```
frickjack: Love this `/flavors/` stuff! :-)

frickjack: Where does `nginx` get configured to proxy kibana?

tf_files/aws/modules/utility-vm/cloud.tf
fauzigo: indeed

fauzigo: change pushed

tf_files/aws/modules/utility-vm/cloud.tf
fauzigo: This is the same AMI that is used for everything else, it comes already loaded with awslogs and it is easier to hook up to cloudwatch. We could copy a public Ubuntu AMI I suppose, I'll check out the outcome os said change today/.

Moreover, if we want to dockerize the service or not, it's got nothing to do with the image itself. Services and packages would be handled by a script in the flavors folder.

PR #319, opened 2018-05-23
fix(warning): remove warning google json

thanh-nguyen-dang:
@frickjack r?

### DIFF
--- a/apis_configs/fence_settings.py
+++ b/apis_configs/fence_settings.py
@@ -74,9 +74,6 @@ def load_json(file_name):
 def get_from_dict(dictionary, key, default=''):
     value = dictionary.get(key)
     if value is None:
-        print(
-            'Warning: A value for key {} not found. Defaulting to "{}"...'
-            .format(key, default))
         value = default
     return value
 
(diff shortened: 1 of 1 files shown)


apis_configs/fence_settings.py
```diff
@@ -113,19 +114,6 @@ def get_from_dict(dictionary, key, default=''):
     HTTP_PROXY = data['HTTP_PROXY']
     dbGaP = data.get('dbGaP',DEFAULT_DBGAP)
 
-    CIRRUS_CFG = {}
```
Avantol13: We still need this configuration if it exists... this will break DCF

Avantol13: You can remove the print statement from get_from_dict if you don't want the warnings

thanh-nguyen-dang: I moved it to line 104 @Avantol13

Avantol13: @thanh-nguyen-dang The entire block of code under this line cannot be deleted. We need to read these configuration variables from the `fence_credentials.json` file. These are all required for fence to operate correctly for Google-related endpoints. This will break existing functionality. If you want to remove the warnings, just remove the print statement in `get_from_dict`

PR #322, opened 2018-05-29
Feat/aws bucket sync

giangbui:


gen3/bin/kube-setup-secrets.sh
```diff
@@ -208,7 +208,25 @@ if [[ -f "${WORKSPACE}/${vpc_name}_output/creds.json" ]]; then  # update secrets
...
+  fi
+
+  if ! g3kubectl get configmaps/data-service-manifest > /dev/null 2>&1; then
+    if [[ ! -f "./apis_configs/manifest" ]]; then
+      touch "./apis_configs/manifest"
+    fi
+    g3kubectl create configmap data-service-manifest --from-file=manifest=./apis_configs/manifest
+  fi
+
   # Create the 'sheepdog' and 'peregrine' postgres user if necessary
```
frickjack: We don't need these secrets and configurations in most commons, so we should only create them if necessary.  Move all these confile file to be under one directory - `apis_configs/dcf_dataservice/creds.json` and `apis_configs/dcf_dataservice/manifest`, and wrap this whole block in an `if` block:
```
if [[ -d apis_configs/dcf_dataservice ]]; then ...
```

kube/services/jobs/aws-bucket-replicate-job.yaml
```diff
@@ -0,0 +1,92 @@
...
+            secretName: "dcf-dataservice-json-secret"
+      containers:
+      - name: awsreplicate
+        image: quay.io/cdis/dcf-dataservice:feat_data_replicate_service
+        imagePullPolicy: Always
+        env:
+          - name: GDC_BUCKET_NAME
+            GEN3_GDC_BUCKET_NAME
+          - name: LOG_BUCKET
+            GEN3_LOG_BUCKET
```
frickjack: What is the LOG_BUCKET for?

kube/services/jobs/aws-bucket-replicate-job.yaml
```diff
@@ -0,0 +1,92 @@
+# run with 
+# g3k runjob aws-bucket-sync \
+# GDC_BUCKET_NAME a_bucket_name \
+# CHUNK_SIZE chunk_size
+#
+# folow the below link  to setup bucket owner grant access to other aws account 
+# https://docs.aws.amazon.com/AmazonS3/latest/dev/example-walkthroughs-managing-access-example2.html
+#
+# setup bucket policy so that the user test can access to the bucket (GDC bucket)
```
frickjack: Would you give some more context here or in jobs/README.md?
Ex - this job is used to copy objects from GDC to an AWS bucket using a manifest ... something like that?

kube/services/jobs/aws-bucket-replicate-job.yaml
giangbui: To save all the uuid which can not be copied.

kube/services/jobs/aws-bucket-replicate-job.yaml
giangbui: fixed at e9fdab74ac2aa33db7836eb27afa2890695ce597

gen3/bin/kube-setup-secrets.sh
giangbui: fixed at e9fdab74ac2aa33db7836eb27afa2890695ce597

PR #330, opened 2018-06-05
Feat/gzip

zflamig:


kube/services/revproxy/00nginx-config.yaml
```diff
@@ -139,7 +139,15 @@ data:
...
-        gzip_types application/javascript;
+        gzip_proxied any;
+        gzip_types
+          text/css
+          text/javascript
+          text/xml
+          text/plain
+          application/javascript
+          application/x-javascript
+          application/json;
```
frickjack: I think we decided not to gzip json due to the BREACH attack:
   https://en.wikipedia.org/wiki/BREACH

kube/services/netpolicy/networkpolicy_fence_templ.yaml
```diff
@@ -23,6 +23,9 @@ spec:
       - podSelector:
           matchLabels:
              app: revproxy
+      - podSelector:
```
frickjack: Why does shiny need to communicate directly with the fence service?
Would it be better if client code accessed our services via our standard public endpoint, than have one way to do things for workspace clients like shiny and another for external clients ?

kube/services/revproxy/00nginx-config.yaml
zflamig: I don't believe we actually reflect user data anywhere so we shouldn't be vulnerable to this?

zflamig: Actually thats not true, we reflect names in graphql queries so you could potentially do it there. Disabling for the graphql endpoint

PR #373, opened 2018-06-20
feat(assume_role): cross AWS accounts

thanh-nguyen-dang:
r @frickjack @philloooo 
related pr:
https://github.com/uc-cdis/fence/pull/257


philloooo:
if the comment in fence is addressed, this change is not needed

PR #374, opened 2018-06-21
Feat/s3 and elb logging csoc

fauzigo:
The way ELB are deployed cannot be tied up with this module, kube-aws is in charge of bringing the up, therefore we must either migrate to ELK or run  commands manually to get the ELB logs into cloudwatch.

As for now this will only send buckets logs into cloudwatch. 

The Environment variable must match a CloudWatch Log group that already exists.


tf_files/aws/modules/s3-bucket/cloud.tf
```diff
@@ -151,3 +151,94 @@ resource "aws_iam_instance_profile" "mybucket_writer" {
...
+resource "aws_iam_role_policy_attachment" "trail_writer_role" {
+  role       = "${aws_iam_role.cloudtrail_writer.name}"
+  policy_arn = "${aws_iam_policy.trail_writer.arn}"
+}
+
+
+# first we need to create a trail in cloudtrail
+
+resource "aws_cloudtrail" "logger_trail" {
+  name                          = "${var.environment}-trail"
```
frickjack: What happens if we have multiple buckets - will they all create the same trail?

tf_files/aws/modules/s3-logs/cloud.tf
```diff
@@ -33,6 +35,13 @@ resource "aws_s3_bucket" "log_bucket" {
   }
 }
 
+#### Added by fauzi@uchicago.edu
+
+data "aws_caller_identity" "current" {}
```
frickjack: Where is this used?

tf_files/aws/modules/s3-bucket/cloud.tf
```diff
@@ -151,3 +151,94 @@ resource "aws_iam_instance_profile" "mybucket_writer" {
...
+  s3_bucket_name                = "${module.cdis_s3_logs.log_bucket_name}"
+  s3_key_prefix                 = "trailLogs"
```
frickjack: So the parent s3-bucket and cloud-trail both write to the logs bucket?

tf_files/aws/modules/s3-logs/cloud.tf
fauzigo: I removed the association with the "data" resource in a previous commit and forgot to remove the invocation.

Removed now

tf_files/aws/modules/s3-bucket/cloud.tf
fauzigo: Yes, 

Regular logging, like the module is doing already, would send logs in plain text to the other S3 bucket. Whereas the trail would send similar information but zipped and there is no way to tell CloudTrail to not deliver zipped files.

From my perspective we have two additional options for this redundancy. We could either set a short lifecycle for either one of those (preferably the plain text ones), or we could just not enable bucket logs other than what goes by CloudTrail.

>CloudTrail tracks API access for infrastructure-changing events, in S3 this means creating, seleting, and modifying bucket (S3 CloudTrail docs). It is very focused on API methods that modify buckets.
S3 Server Access Logging provides web server-style logging of access to the objects in an S3 bucket. This logging is granular to the object, includes read-only operations, and includes non-API access like static web site browsing.

tf_files/aws/modules/s3-bucket/cloud.tf
fauzigo: I wasn't counting on having multiple buckets for the same commons, but it makes sense to have the module ready for multiple buckets.

fauzigo: pushed a change that should take care of this in the future.

PR #379, opened 2018-06-26
Centralised squid nlb set-up in CSOC

rarya:
This branch have the code for spinning up the centralised squid nlb set-up in CSOC.
It also has code for creating an endpoint in the commons (pointing to the squid nlb endpoint service in CSOC) and the code for creating the DNS entry for cloud-proxy (pointing to the DNS of the endpoint in commons). However, the code for creating the DNS entry for the cloud-proxy  in commons (pointing to centralised set-up) is commented out for now.


flavors/squid_nlb_central/squidvm.sh
```diff
@@ -0,0 +1,154 @@
+#!/bin/bash
+
+SUB_FOLDER="/home/ubuntu/cloud-automation/"
+MAGIC_URL="http://169.254.169.254/latest/meta-data/"
+
+
+cd /home/ubuntu
+sudo apt-get update
+sudo apt-get install -y build-essential wget libssl-dev
+wget http://www.squid-cache.org/Versions/v4/squid-4.0.24.tar.xz
```
frickjack: Would you add some instructions on where to go to check for new versions, and a note on why we're running v4 before it goes GA, and that we'll switch to the `apt` package once v4 is available in the standard repos?

flavors/squid_nlb_central/updatewhitelist.sh
```diff
@@ -0,0 +1,36 @@
...
+
+cd /home/ubuntu/cloud-automation
+git pull
+
+DIFF_AUTH1=$(diff "/home/ubuntu/cloud-automation/flavors/squid_nlb_central/authorized_keys_admin" "/home/ubuntu/.ssh/authorized_keys")
+DIFF_AUTH2=$(diff "/home/sftpuser/cloud-automation/flavors/squid_nlb_central/authorized_keys_user" "/home/sftpuser/.ssh/authorized_keys")
+
+DIFF_SQUID1=$(diff "/home/ubuntu/cloud-automation/flavors/squid_nlb_central/web_whitelist" "/etc/squid/web_whitelist")
+DIFF_SQUID2=$(diff "/home/ubuntu/cloud-automation/flavors/squid_nlb_central/web_wildcard_whitelist" "/etc/squid/web_wildcard_whitelist")
+DIFF_SQUID3=$(diff "/home/ubuntu/cloud-automation/flavors/squid_nlb_central/ftp_whitelist" "/etc/squid/ftp_whitelist")
```
frickjack: It will be better to just check the exit code from `diff` to see if something has changed - ex:
```
echo "Checking if web_whitelist has changed"
diff ...
DIFF_SQUID1=$?
if [[ $DIFF_SQUID1 -ne 0 ]]; then ... the file has changed 
```

gen3/lib/aws.sh
```diff
@@ -280,6 +283,22 @@ EOM
...
+    # rds snapshot vpc is simpler ...
+    commonsName=$(echo "$GEN3_WORKSPACE" | sed 's/_snapshot$//')
+    cat - <<EOM
+  env_vpc_octet3                = "3rd OCTET OF CSOC CIDR FOR SQUID SETUP"
+  env_nlb_name                  = "NLB SETUP NAME"
+  env_vpc_id                    = "CSOC VPC-ID"
+  env_pub_subnet_routetable_id = "CSOC ROUTE TABLE ID - HAVING ROUTE TO INTERNET GW"
+  csoc_internal_dns_zone_id  = "PUT IT AS `ZA1HVV5W0QBG1` IF LAUNCHING THE SQUID NLB IN CSOC MAIN VPC"
+  allowed_principals_list       = "[LIST OF AWS ACCOUNTS WHICH NEEDS TO BE WHITELISTED]"
+  # e.g. of the list - ["arn:aws:iam::<AWS ACCOUNT1 ID>:root","arn:aws:iam::<AWS ACCOUNT2 ID>:root", ...]
```
frickjack: Can we put real default values in here for our current CSOC?  We'll only be running one central squid for now, right?  :-)

tf_files/aws/commons/variables.tf
```diff
@@ -133,3 +133,6 @@ locals {
   # kube-aws does not like '-' in cluster name
   cluster_name = "${replace(var.vpc_name, "-", "")}"
 }
+
+variable "squid-nlb-endpointservice-name" {
+  }
```
frickjack: We should have a default value here that points at the CSOC squid service?

tf_files/aws/modules/squid_nlb_central_csoc/README.md
```diff
@@ -0,0 +1,5 @@
+# Description
+A centralised NLB set-up for Squid proxy will be installed in the CSOC account. This is going to be a centralised squid set-up utilized by all the data commons VPC.
+
+   
+
```
frickjack: Let's make a lucid chart SVG, and link it here:
* CSOC
   - 3 subnets
   - autoscaling group
   - NLB 
   - endpoint service?  is that a thing?
* Child VPC
    - vpc endpoint
    - note about cloud-proxy.internal.io DNS entry

Also put a note about put a note about how we cannot point a default route at the endpoint ...

frickjack: Also put a note about pricing of bandwidth through the VPC endpoint

flavors/squid_nlb_central/squidvm.sh
rarya: done

flavors/squid_nlb_central/updatewhitelist.sh
rarya: done.

gen3/lib/aws.sh
rarya: I think we better keep it this way, allowed_principal list will be modified whenever an account is added.Anyways this is launched from the csoc_master_admin VM; which anyways have all those valuse in the `.local/share/gen3/csoc/csocsquidnlb_squidnlbcentral` config.tvars file.

tf_files/aws/commons/variables.tf
rarya: done.

PR #384, opened 2018-06-27
chore(squidnlb: changes to the updatewhitelish.sh and commons code  t…

rarya:
…o add the vpc endpoint config.tvars variable

### DIFF
--- a/flavors/squid_nlb_central/updatewhitelist.sh
+++ b/flavors/squid_nlb_central/updatewhitelist.sh
@@ -4,7 +4,7 @@ cd /home/ubuntu/cloud-automation
 git pull
 
 echo "Checking if list of authorized keys have changed for admin"
-sudo diff "/home/ubuntu/cloud-automation/flavors/squid_nlb_central/authorized_keys_admin" "/home/ubuntu/.ssh/authorized_keys"
+diff "/home/ubuntu/cloud-automation/flavors/squid_nlb_central/authorized_keys_admin" "/home/ubuntu/.ssh/authorized_keys"
 DIFF_AUTH1=$?
 if [ "$DIFF_AUTH1" -ne 0  ] ; then
 echo "There is a change in authorized_keys for admin"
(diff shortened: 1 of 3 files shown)


flavors/squid_nlb_central/updatewhitelist.sh
```diff
@@ -4,7 +4,7 @@ cd /home/ubuntu/cloud-automation
 git pull
 
 echo "Checking if list of authorized keys have changed for admin"
-diff "/home/ubuntu/cloud-automation/flavors/squid_nlb_central/authorized_keys_admin" "/home/ubuntu/.ssh/authorized_keys"
+sudo diff "/home/ubuntu/cloud-automation/flavors/squid_nlb_central/authorized_keys_admin" "/home/ubuntu/.ssh/authorized_keys"
```
frickjack: Why do you need `sudo` here, but not in the rsync?

gen3/lib/aws.sh
```diff
@@ -194,7 +194,7 @@ vpc_name="$GEN3_WORKSPACE"
 #
 vpc_octet2=GET_A_UNIQUE_VPC_172_OCTET2
 vpc_octet3=GET_A_UNIQUE_VPC_172_OCTET3
-squid-nlb-endpointservice-name=CSOC_SQUID_NLB_SERVICE_NAME
+squid-nlb-endpointservice-name="CSOC_SQUID_NLB_SERVICE_NAME"
```
frickjack: This hides the variable default in `variables.tf` - put the "com.amazonaws.vpce.us-east-1.vpce-svc-0ce2261f708539011" value here too, or just don't put the variable in `config.tfvars`

flavors/squid_nlb_central/updatewhitelist.sh
```diff
@@ -4,7 +4,7 @@ cd /home/ubuntu/cloud-automation
 git pull
 
 echo "Checking if list of authorized keys have changed for admin"
```
frickjack: Does this script run as the `ubuntu` user in a crontab?  If so - please add a comment explaining that.

flavors/squid_nlb_central/updatewhitelist.sh
```diff
@@ -22,23 +22,23 @@ fi
```
frickjack: If this script runs as the `ubuntu` user, then I think these `rsync` commands need to be run with `sudo` too

flavors/squid_nlb_central/updatewhitelist.sh
rarya: the cronjob is run as root; so removed sudo

gen3/lib/aws.sh
rarya: I removed it from the config.tvars

PR #388, opened 2018-06-29
Feat/google bucket replicate

giangbui:
cloud automation job for google bucket replication using google data flow


zflamig:
I know we merged the AWS sync already, but since this is really project specific it should probably live elsewhere. WDYT @frickjack?

PR #395, opened 2018-07-05
feat(path): support yaml full path

giangbui:


gen3/lib/g3k.sh
```diff
@@ -138,8 +138,13 @@ g3k_runjob() {
     echo "g3k runjob JOBNAME"
     return 1
   fi
-  jobPath="${GEN3_HOME}/kube/services/jobs/${jobName}-job.yaml"
-  if [[ -f "$jobPath" ]]; then   
+  jobPath="$jobName"
+  if [[ -f "$jobPath" ]]; then
+      jobName="$(basename $jobPath | sed 's/.yaml$//')"
+  else
```
frickjack: Small change - add " and also trim '-job':
```
jobName="$(basename "$jobPath" | sed 's/-job.yaml$//')"
```

PR #397, opened 2018-07-05
PXD-824 Feat(compose integration): integrating local docker compose setup with gen3

stevencui729:


compose-files/creds_setup.sh
```diff
@@ -0,0 +1,77 @@
+#!/bin/bash
```
frickjack: Would you move this to 'Docker/compose-files` rather than setup a root level directory?

compose-files/docker-compose.yml
```diff
@@ -0,0 +1,98 @@
+version: '3'
+services:
+  indexd:
+    image: "quay.io/cdis/indexd"
```
frickjack: `:master` ?

compose-files/docker-compose.yml
```diff
@@ -0,0 +1,98 @@
...
+    volumes:
+      - ./apis_configs/indexd_settings.py:/var/www/indexd/local_settings.py
+      - ./apis_configs/creds.json:/var/www/indexd/creds.json
+      - ./apis_configs/config_helper.py:/var/www/indexd/config_helper.py
+      - ./waitForPostgres.sh:/var/www/indexd/waitForPostgres.sh
+    depends_on:
+      - postgres
+  fence:
+    image: "quay.io/cdis/fence"
+    networks:
```
frickjack: `:master` image - tag all the images `:master`

tf_files/shared/modules/compose_configs/creds.tpl
```diff
@@ -0,0 +1,47 @@
+{
```
frickjack: Don't put this here.  These are files for terraform.  Put everything compose related under `Docker/compose-files`

compose-files/docker-compose.yml
```diff
@@ -0,0 +1,98 @@
...
+      - devnet
+    volumes:
+      - ./apis_configs/fence_settings.py:/var/www/fence/local_settings.py
+      - ./apis_configs/creds.json:/var/www/fence/creds.json
```
frickjack: Where do you extract the `indexd` block from `creds.json` ?  The `config_helper` and `settings.py` files expect the mounted `creds.json` to just have the block appropriate for the running service

stevencui729: Oh my workaround for the `config_helper` and `settings.py` files expecting just the service-specific block was to change the code in the `settings.py` files to access `creds.json` by key. In the compose-services repo, the `settings.py` files are directly changed, but here in cloud automation I used sed to modify the lines

PR #406, opened 2018-07-16
Feat/es module

fauzigo:


gen3/bin/kube-setup-aws-es-proxy.sh
```diff
@@ -0,0 +1,22 @@
...
+    exit 1
+fi
+
+source "${GEN3_HOME}/gen3/lib/utils.sh"
+gen3_load "gen3/lib/kube-setup-init"
+
+gen3 kube-setup-secrets
+#gen3 roll esproxy
+sed "s/ES_ENDPOINT/https:\/\/${1}/" "${GEN3_HOME}/kube/services/aws-es-proxy/aws-es-proxy-deploy.yaml" | g3kubectl apply -f -
+g3kubectl apply -f "${GEN3_HOME}/kube/services/netpolicy/networkpolicy_aws_es_proxy.yaml"
```
frickjack: So this expects the ES-PROXY endpoint to be passed as a commond line argument to `kube-setup-aws-es-proxy`?

gen3/bin/kube-setup-secrets.sh
```diff
@@ -28,6 +28,20 @@ if [[ -f "${WORKSPACE}/${vpc_name}/creds.json" ]]; then # update secrets
...
+  if ! g3kubectl get secrets/aws-es-proxy > /dev/null 2>&1; then
+    credsFile=$(mktemp -p "$XDG_RUNTIME_DIR" "creds.json_XXXXXX")
+    creds=$(jq -r ".es|tostring" < creds.json |sed -e 's/[{-}]//g' -e 's/"//g' -e 's/:/=/g')
+    echo "[default]" > "$credsFile"
+    IFS=',' read -ra CREDS <<< "$creds"
+    for i in "${CREDS[@]}"; do
+      echo ${i} >> "$credsFile"
+    done
+    g3kubectl create secret generic aws-es-proxy "--from-file=credentials=${credsFile}"
+  fi
```
frickjack: Can we add the ES_PROXY endpoint to the secret too, then just mount that into the container rather than run `sed` on the deploy.yaml ?

frickjack: Also - we should probably leave a copy of the secret under `${vpc_name}/apis_configs/es/` - something like that?  It's not obvious to me where `creds.json` is coming from - maybe you meant to use `$credsFile` ?  I can stop by to have you explain it to me :-p

gen3/lib/aws.sh
```diff
@@ -322,6 +324,16 @@ EOM
     return 0
   fi
 
+  # else ...
+  if [[ "$GEN3_WORKSPACE" =~ _es$ ]]; then
+      commonsName=${GEN3_WORKSPACE//_es/}
+      cat - <<EOM
```
frickjack: Would you verify the `//_es/` trick works in `zsh`?  Zsh drives me crazy ...

kube/services/aws-es-proxy/aws-es-proxy-deploy.yaml
```diff
@@ -0,0 +1,42 @@
...
+        - containerPort: 9200
+        volumeMounts:
+          - name: "credentials"
+            readOnly: true
+            mountPath: "/root/.aws"
+        command: ["./aws-es-proxy"]
+        args:
+          - "-endpoint"
+          - "ES_ENDPOINT"
+          - "-verbose"
```
frickjack: Yeah - pull ES_ENDPOINT in from a secret or configmap - could even put it into the global configmap if you want ...

kube/services/netpolicy/networkpolicy_aws_es_proxy.yaml
```diff
@@ -0,0 +1,23 @@
...
+      app: esproxy
+  ingress:
+    - from:
+      - podSelector:
+          matchLabels:
+            app: arranger
+      ports:
+         - port: 80
+         - port: 443
+  egress:
```
frickjack: How about we go ahead and public to port `9200` - since that's where everybody expects Elastic Search to be anyway?  And go ahead and get rid of port 443 ...  For sure we need a `targetPort: 9200` ...?

Actually - this is the `networkpolicy` - don't we need an `esproxy-service` too?

tf_files/aws/modules/commons-vpc-es/cloud.tf
```diff
@@ -0,0 +1,126 @@
...
+            "Resource": "*"
+        }
+    ]
+}
+CONFIG
+}
+
+data "aws_iam_user" "es_user" {
+  user_name = "${var.vpc_name}_es_user"
+}
```
frickjack: I was jut thinking - is it possible to setup an `es_role` with permissions to access ES, and an `es_user` with permission to assume the role, and configure the `es-proxy` to assume the role to access ES?  If `es-proxy` doesn't support that, then forget it, but that might be a nice setup if we ever switch over to getting creds from the metadata service ...

gen3/bin/kube-setup-aws-es-proxy.sh
fauzigo: Yes, I guess I could try querying with awscli, going to try something out.

This is what's outputed by tfapply, you would need the es_endopoint.

```
Apply complete! Resources: 4 added, 0 changed, 0 destroyed.

Outputs:

es_endpoint = vpc-fauziv1-gen3-metadata-kydadmgkzectkhr2pdd2s2uihu.us-east-1.es.amazonaws.com
kibana_endpoint = vpc-fauziv1-gen3-metadata-kydadmgkzectkhr2pdd2s2uihu.us-east-1.es.amazonaws.com/_plugin/kibana/
Backing up files to s3
Backing up config.tfvars to s3://cdis-state-ac707767160287-gen3/fauziv1_es/config.tfvars
upload: ./config.tfvars to s3://cdis-state-ac707767160287-gen3/fauziv1_es/config.tfvars
Backing up backend.tfvars to s3://cdis-state-ac707767160287-gen3/fauziv1_es/backend.tfvars
upload: ./backend.tfvars to s3://cdis-state-ac707767160287-gen3/fauziv1_es/backend.tfvars
Backing up README.md to s3://cdis-state-ac707767160287-gen3/fauziv1_es/README.md
upload: ./README.md to s3://cdis-state-ac707767160287-gen3/fauziv1_es/README.md
```

gen3/lib/aws.sh
fauzigo: my environment in cdistest is zsh and it works just fine. Also tested it locally before using it.

So, yeah, it works just fine,.

Issue #415, opened 2018-07-23
PXD-1224 ⁃ Launching a Qualys VM via Terraform


rarya:
Working with Jeff to automate the process of launching a Qualys scanner in CSOC.

PR #420, opened 2018-07-26
Automation for qualys scanner VM in csoc

rarya:
#415


gen3/lib/aws.sh
```diff
@@ -334,6 +336,16 @@ EOM
   fi
 
 
+  # else ...
+  if [[ "$GEN3_WORKSPACE" =~ _qualysvm$ ]]; then
+      commonsName=${GEN3_WORKSPACE//_es/}
```
fauzigo: Are you sure that's what you want to replace? shouldn't it be 
`commonsName=${GEN3_WORKSPACE//_qualysvm/}` instead

PR #427, opened 2018-08-06
Feat/kube lock

stevencui729:


gen3/bin/kube-lock.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
+#
+# Script to lock namespace using labels on a lock ConfigMap (make sure 
+# KUBECTL_NAMESPACE is set so that g3kubectl works properly)
+
+help() {
+  cat - <<EOM
+  gen3 kube-lock lock-name owner:
+    Attempts to lock the lock lock-name in the namespace that KUBECTL_NAMESPACE
```
frickjack: Can we add a timeout to a lock - so a lock auto-expires/unlocks - something like:
   `gen3 kube-lock lock-name owner max-age`

gen3/bin/kube-lock.sh
```diff
@@ -0,0 +1,59 @@
...
+EOM
+  return 0
+}
+
+if [[ $1 =~ ^-*help$ || $# -ne 2 ]]; then
+  help
+  exit 0
+fi
+
+set -i
```
frickjack: Would be cool to alias $1 and $2 - just to make things easier to read - ex:
```
lockName="$1"
owner="$2"
```

PR #446, opened 2018-08-26
feat(user_registration): add job and cronjob for google manage user

giangbui:


kube/services/jobs/google-manage-user-registrations-cronjob.yaml
```diff
@@ -0,0 +1,85 @@
...
+                secretName: "fence-google-storage-creds-secret"
+            - name: fence-yaml
+              configMap:
+                name: fence
+            - name: config-helper
+              configMap:
+                name: config-helper
+          containers:
+          - name: fence
+            image: quay.io/cdis/fence:master
```
philloooo: can we use `GEN3_FENCE_IMAGE` ?

PR #461, opened 2018-09-11
Fix/arborist

philloooo:
- without the networkpolicy it still works in dev, so we should fix the dev environment to respect networkpolicy so testing in dev is more useful
- added some doc

### DIFF
--- a/kube/services/netpolicy/networkpolicy_arborist_templ.yaml
+++ b/kube/services/netpolicy/networkpolicy_arborist_templ.yaml
@@ -1,0 +1,39 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: networkpolicy-arborist
+spec:
+  podSelector:
+    matchLabels:
+      app: arborist
+  ingress:
+    - from:
+      - podSelector:
+          matchLabels:
+            app: arborist
+      - podSelector:
+          matchLabels:
+             app: revproxy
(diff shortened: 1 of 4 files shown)


kube/services/netpolicy/networkpolicy_arborist_templ.yaml
```diff
@@ -0,0 +1,39 @@
...
+            app: arborist
+      - podSelector:
+          matchLabels:
+            app: fence
+      - ipBlock:
+           cidr: 0.0.0.0/0
+      - ipBlock:
+           cidr: 54.224.0.0/12
+      - ipBlock:
+           cidr: 52.92.16.0/20
```
frickjack: Does arborist need egress to the internet?

philloooo: oh I guess not yet

frickjack:
you need to deploy some network policy to your dev environment for it to take affect: `gen3 kube-setup-networkpolicy` or `gen3 roll all` should do it too - if no network policies have been deployed, then the cluster assumes open communication

frickjack:
Ugh - maybe I spoke too soon - might need to deploy Calico on the cluster - I'll talk with Fauzi:
   https://docs.aws.amazon.com/eks/latest/userguide/calico.html

philloooo:
Yeah it was a new cluster so I rolled all

PR #465, opened 2018-09-12
Feat/g3k mani2cmap: New configmap creation and tests

stevencui729:


gen3/lib/g3k.sh
```diff
@@ -271,6 +271,52 @@ g3k_ec2_reboot() {
...
+
+  local key
+  local key2
+  local value
+  local execString
+
+  for key in $(g3k_config_lookup 'keys[]' "$manifestPath"); do
+    if [[ $key != 'notes' ]]; then
+      cMapName="manifest-$key"
+      execString="g3kubectl create configmap $cMapName "
```
frickjack: add `local cMapName` ?

frickjack: also `local jsonSection` ...

PR #467, opened 2018-09-13
chore(update-minit): added branch specification to g3k_manifest_init

stevencui729:


stevencui729:
Sure, I can do that. The environment variable sounds really cool, that sounds like it should be the primary way we use this functionality

PR #479, opened 2018-09-21
Fix/workers ami eks

fauzigo:


fauzigo:
The new AMI released by amazon is kind of buggy.

New images would ship with a bootstrap script that is supposed to do the necessary changes and installs we currently do in user_data when we configure the autoscaling group. However, said script at some point needs to run aws commands to talk to EKS and obtain information about it, but `aws eks ...` would just fail and the script exits.

If we can not use the AMI directly without tweaking it, there is no reason to update as of now.

PR #481, opened 2018-09-25
PXD-1896 feat(google_links): force google links

giangbui:


gen3/bin/kube-setup-secrets.sh
```diff
@@ -227,6 +227,13 @@ if [[ -f "${WORKSPACE}/${vpc_name}/creds.json" ]]; then # update fence secrets
   fi
 fi
 
+if ! g3kubectl get configmaps/isb-cgc-link-list > /dev/null 2>&1; then
+  if [[ ! -f "./apis_configs/isb-cgc-link-list.txt" ]]; then
+    touch "apis_configs/isb-cgc-link-list.txt"
```
Avantol13: we should not use their name in our code, need to call this something else. external-google-link-list or something

gen3/bin/kube-setup-secrets.sh
```diff
@@ -227,6 +227,13 @@ if [[ -f "${WORKSPACE}/${vpc_name}/creds.json" ]]; then # update fence secrets
   fi
 fi
 
+if ! g3kubectl get configmaps/isb-cgc-link-list > /dev/null 2>&1; then
+  if [[ ! -f "./apis_configs/isb-cgc-link-list.txt" ]]; then
```
Avantol13: this will get created in every environment. I think you should just change cloud automation to this branch to run this when we need to but we shouldn't actually merge this code since this is not something we should ever have to do again.

PR #521, opened 2018-10-16
feat(tube-doc): document for tube

thanh-nguyen-dang:


frickjack:
Also - is the mapping file documented any place?  If so - add a link to the docs.  If not, then please document the format of the mapping file, and give a sample mapping. :+1:

PR #551, opened 2018-10-25
feat(mock_login): add MOCK_GOOGLE_AUTH config

giangbui:


apis_configs/fence_settings.py
```diff
@@ -153,6 +153,8 @@ def get_from_dict(dictionary, key, default=""):
     REMOVE_SERVICE_ACCOUNT_EMAIL_NOTIFICATION = data.get(
         "REMOVE_SERVICE_ACCOUNT_EMAIL_NOTIFICATION"
     )
+    # use for intergration tests to skip the login page
+    MOCK_GOOGLE_AUTH = data.get("MOCK_GOOGLE_AUTH")
```
Avantol13: just for absolute clarity you should default it to `False` instead of `None`.
```suggestion
    MOCK_GOOGLE_AUTH = data.get("MOCK_GOOGLE_AUTH", False)
```

PR #554, opened 2018-10-25
Chore/keys better handling

fauzigo:


tf_files/aws/modules/eks/root.tf
```diff
@@ -20,4 +20,7 @@ data "template_file" "init_cluster" {
     auth_configmap  = "${var.vpc_name}_output_EKS/aws-auth-cm.yaml"
   }
 }
-    
+ 
+data "template_file" "ssh_keys" {
+  template = "${file("~/cloud-automation/files/authorized_keys/ops_team")}"
```
frickjack: Can you make this a path relative to the module like this? :
   https://github.com/uc-cdis/cloud-automation/blob/master/tf_files/shared/modules/k8s_configs/root.tf#L2

PR #580, opened 2018-11-14
Feat/upload data bucket PXD-2215

fauzigo:

### New Features
- New databucket creation (<vpc_name>-data-bucket) when the commons resources are initially spun up, right before deploying the kubernetes cluster (EKS). Also a new fence-bot (<vpc_name>-fence-bot) user is created, i can be used or just ignored if there is already an active user. 

Additionally, actions against the databucket are stored in another bucket (<vpc_name>-data-bucket-logs) and in cloudwatch, which would also send them over to CSOC through the kinesis stream.

### Breaking Changes
N/A

### Bug Fixes
N/A

### Improvements
No manually created users for databucket should be done from now on

### DIFF
--- a/tf_files/aws/modules/upload-data-bucket/s3.tf
+++ b/tf_files/aws/modules/upload-data-bucket/s3.tf
@@ -1,0 +1,107 @@
+
+## The actual data bucket
+
+resource "aws_s3_bucket" "data_bucket" {
+  bucket = "${var.vpc_name}-data-bucket"
+  acl    = "private"
+
+  server_side_encryption_configuration {
+    rule {
+      apply_server_side_encryption_by_default {
+        sse_algorithm = "AES256"
+      }
+    }
+  }
+
+  logging {
+    target_bucket = "${aws_s3_bucket.log_bucket.id}"
+    target_prefix = "log/${var.vpc_name}-data-bucket"
+  }
+
+  tags {
+    Name        = "${var.vpc_name}-data-bucket"
+    Environment = "${var.environment}"
+    Purpose     = "data bucket"
+  }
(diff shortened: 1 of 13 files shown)


tf_files/aws/modules/upload-data-bucket/s3.tf
```diff
@@ -0,0 +1,107 @@
...
+
+    prefix = "/"
+
+    tags {
+      "rule"      = "log"
+      "autoclean" = "true"
+    }
+
+    expiration {
+      days = 120
```
frickjack: we expire the logs here, because we know they'll also be logged in cloud-trail?

tf_files/aws/modules/vpc/cloud.tf
```diff
@@ -10,6 +10,20 @@ module "squid_proxy" {
...
+  cloudwatchlogs_group = "${aws_cloudwatch_log_group.main_log_group.arn}"
+  environment          = "${var.vpc_name}"
+}
+
+module "fence-bot-user" {
+  source               = "../fence-bot-user"
+  vpc_name             = "${var.vpc_name}"
+  bucket_name          = "${module.data-bucket.data-bucket_name}"
+}
+
```
frickjack: this is awesome :+1:

tf_files/aws/modules/upload-data-bucket/s3.tf
fauzigo: Basically yes, they are also being sent over to CSOC, I don't think we necessarily need them redundant on both sides (commons account and CSOC), but this is and easy fix if we need them both up to date

PR #595, opened 2018-11-21
chore(creds): add indexd password to fence creds

rudyardrichter:
Add `indexd_password` field to the fence `creds.json`

### DIFF
--- a/tf_files/shared/modules/k8s_configs/creds.tpl
+++ b/tf_files/shared/modules/k8s_configs/creds.tpl
@@ -5,7 +5,7 @@
         "db_password": "${fence_pwd}",
         "db_database": "${fence_db}",
         "hostname": "${hostname}",
-        "indexd_password": "${gdcapi_indexd_password}",
+        "indexd_password": "",
         "google_client_secret": "${google_client_secret}",
         "google_client_id": "${google_client_id}",
         "hmac_key": "${hmac_encryption_key}"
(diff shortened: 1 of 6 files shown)


tf_files/shared/modules/k8s_configs/creds.tpl
```diff
@@ -5,6 +5,7 @@
         "db_password": "${fence_pwd}",
         "db_database": "${fence_db}",
         "hostname": "${hostname}",
+        "indexd_password": "${gdcapi_indexd_password}",
```
Avantol13: this indexd_password is for the gdcapi user in indexd? So it seems like we're just using that user for basic auth for all our microservices? this has always been kinda confusing to me, I don't think it's within the scope of this, but perhaps we should have different users for each microservice and get rid of all this gdcapi stuff?

Also, this is not going to work for existing commons. The creds.json file only gets generated once when the commons is spun up with terraform. so you're going to need to add some logic to update existing creds.json with this field.

I'm not really sure where in cloud auto something like that should go... probably in gen3/bin/kube-setup-fence (you should verify this with reuben and maybe have him review too)?

I did a quick search to see if there's anywhere else creds.json gets update. found: https://github.com/uc-cdis/cloud-automation/blob/master/gen3/bin/kube-dev-namespace.sh#L115-L125 code might be useful, that's not where I'd add your changes though.

PR #636, opened 2018-12-31
Feat/management logs PXD-2306

fauzigo:

### New Features
- Management Logging.

Introducing a couple of new modules; management-logs and account_management-logs. 

These modules are per account module, not per commons. The first one is to be ran ONLY on the csoc account and also only once. The second one is for all other accounts.

The main idea is to send all management logs collected by cloud trail to the CSOC account.

Logs will be stored in S3 with a lifecycle of after 80 days will be sent to Glacier, and these one kept for 5 years before expiring.


tf_files/aws/modules/management-logs/logging.tf
```diff
@@ -0,0 +1,387 @@
...
+
+# Let's not use the zip file and have teraraform zip it for us on the fly
+
+data "archive_file" "lambda_function" {
+  type        = "zip"
+  source_file = "${path.module}/lambda_function.py"
+  output_path = "lambda_function_payload.zip"
+}
+
+resource "aws_lambda_function" "logs_decodeding" {
```
frickjack: spelling `logs_decodeding` ....

tf_files/aws/modules/management-logs/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+
+
+## 3. Centralized logging
+
+For each account, all management logs collected by cloud-trail will be send to CloudWatchLogs
+
+This CSOC account will also collect all those logs generated by its child accounts, for a more easy way to browse them all if required without needing to log into each account.
+
+As for now, logs will be send to a bucket. We can later send them to ES if we wanted to.
```
frickjack: It looks like the logs are being set to ES too?  Can we make that optional in the terraform code based on input variables?

tf_files/aws/modules/account-management-logs/cloud.tf
```diff
@@ -0,0 +1,175 @@
...
+    Organization = "CTDS"
+  }
+}
+
+
+# Finally the subscription filter
+
+resource "aws_cloudwatch_log_subscription_filter" "csoc_subscription" {
+  name            = "${var.account_name}_subscription"
+  destination_arn = "arn:aws:logs:${data.aws_region.current.name}:${var.csoc_account_id}:destination:management-logs_logs_destination"
```
frickjack: Why not make the entire `destination_arn` a variable rather than require that the destination account follow a naming convention?
Also - can we make the subscription_filter optional? `count = ...`

PR #639, opened 2019-01-10
Feat/nodepooling PXP-1980

fauzigo:

### New Features
- Fully separated worker pool for jupyterhub

### Improvements
- This should keep jupyter workers on their own ASG with their own role.
- This should also mitigate the issue where the pods take longer to respond requests of services.


files/scripts/jupyter_migration.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
+
+
+# We need to list all the existing volumes related to Jupyter, usually named pvc-"[Aa0-zZ9]?-*", and we have to rule out the one for the deployment.
+# Then we should create snapshots for those volumens 
+# kubectl get persistentvolume --kubeconfig fauziv1/kubeconfig-KUBE_AWS -o name |cut -d/ -f2 |grep -v pvc-5a172428-d643-11e8-82ba-0e3bbdfa5f60 |while read i; do aws ec2 create-snapshot --description "SS-${i}" --tag-specifications "ResourceType=snapshot,Tags=[{Key=Name,Value=${i}}]" --volume-id $(kubectl --kubeconfig fauziv1/kubeconfig-KUBE_AWS get persistentvolume -o yaml --export $i |yq -r .spec.awsElasticBlockStore.volumeID | awk -F/ '{print $NF}'); done
+
+# kubectl get persistentvolume -o name |cut -d/ -f2 |grep -v pvc-47b422cc-854a-11e8-84de-0ec14f5e5db4 |while read i; do aws ec2 create-snapshot --description "SS-${i}" --tag-specifications "ResourceType=snapshot,Tags=[{Key=Name,Value=${i}}]" --volume-id $(kubectl get persistentvolume -o yaml --export $i |yq -r .spec.awsElasticBlockStore.volumeID | awk -F/ '{print $NF}'); done
+
+AZs=("a" "c" "d")
```
frickjack: why are AZs hard coded to a, c, d?

files/scripts/jupyter_migration.sh
```diff
@@ -0,0 +1,65 @@
...
+COUNTER=0
+
+# When we have snapshots of the active volumes, we must restore them so the hub can find them respectively.
+SNAPSHOTS=$(aws ec2 describe-snapshots --filters "Name=description,Values=SS*" --query "Snapshots[].SnapshotId" --output text)
+
+for SNAP in ${SNAPSHOTS};
```
frickjack: So we're scanning a bunch of snapshots, creating a volume from each, and a persistent-volume.yaml for each?

files/scripts/jupyter_migration.sh
```diff
@@ -0,0 +1,65 @@
...
+do
+        # Get the volumenID for which the snapshot belongs to so we can gatther a little more information about it.
+        VOL=$(aws ec2 describe-snapshots --filters "Name=snapshot-id,Values=${SNAP}" --query "Snapshots[].VolumeId" --output text)
+        OLDIFS=${IFS}
+        taggo=""
+        IFS=$'\n'
```
frickjack: should set
```
PVNAME=""
```

files/scripts/jupyter_migration.sh
```diff
@@ -0,0 +1,65 @@
...
+                #then
+                #        PVCNAME="${VAL}"
+                fi
+                taggo="$taggo,{Key=${KEY},Value=${VAL}}"
+        done
+
+
+        taggo=${taggo#","}
+
+        echo "Creating Volume off Snapshot ${SNAP}"
```
frickjack: Add a guard
```
if [[ -n "$PVNAME" ]]; then ...
```

files/scripts/jupyter_migration.sh
fauzigo: The script is not something that you would run and it'll migrate the persistent volumens on it's own, it requires some manual intervention prior running and most likely some minor tweaks.

I'll try to find out a way to fully automate this, but if otherwise, the whole idea is not to run this blindly.

tf_files/aws/modules/eks-nodepool/cloud.tf
```diff
@@ -0,0 +1,342 @@
...
+  description = "security group that only enables ssh"
+  vpc_id      = "${data.aws_vpc.the_vpc.id}"
+
+  ingress {
+    from_port   = 22
+    to_port     = 22
+    protocol    = "TCP"
+    cidr_blocks = ["0.0.0.0/0"]
+  }
+
```
frickjack: If I'm running a jupyter notebood in a pod on the cluster, can I ssh from there into a node?  No, because I still need to have a valid ssh key ...

tf_files/aws/modules/eks-nodepool/cloud.tf
```diff
@@ -0,0 +1,342 @@
...
+  image_id                    = "${data.aws_ami.eks_worker.id}"
+  instance_type               = "${var.instance_type}"
+  name_prefix                 = "eks-${var.vpc_name}-nodepool-${var.nodepool}"
+  security_groups             = ["${aws_security_group.eks_nodes_sg.id}", "${aws_security_group.ssh.id}"]
+  user_data_base64            = "${base64encode(data.template_file.bootstrap.rendered)}"
+  key_name                    = "${var.ec2_keyname}"
+
+  root_block_device {
+    volume_size = 30
+  }
```
frickjack: We should check with Zac or Phillis about this.  I'm not sure how the `jupyter-fuse` stuff works exactly - if an app starts reading an S3 object, does the whole object get copied down locally, or does it just copy down a block at a time on the fly, and never actually stores the whole object locally?

files/scripts/jupyter_migration.sh
```diff
@@ -0,0 +1,327 @@
...
+    done
+}
+
+function destroyJupyter()
+{
+    if [[ "${DESTROY_JUPYTER}" == "YES" ]];
+    then
+        OIFS=${IFS};IFS=$'\n';
+        #for i in $(grep apply ${GEN3_HOME}/gen3/bin/kube-setup-jupyterhub.sh |awk '{print $NF}'); 
+        for i in $(grep apply ${GEN3_HOME}/gen3/bin/kube-setup-jupyterhub.sh |sed  's/apply/delete/')
```
frickjack: this kind of trickery makes me uncomfortable ...

tf_files/aws/modules/eks-nodepool/cloud.tf
fauzigo: Workers do not store any private key and if you can upload one into the notebook, you still need the public one in the node you are trying to ssh in.

PR #674, opened 2019-02-12
Feat/healthcheck PXP-2397

macintoshpie:
gen3 helper for checking status of the cluster

### New Features
`gen3 healthcheck`
Returns info about unhealthy pods and nodes

### DIFF
--- a/gen3/bin/healthcheck.sh
+++ b/gen3/bin/healthcheck.sh
@@ -1,5 +1,5 @@
 #
-# TODO: write info here
+# Finds any unhealthy pods and nodes
 #
 
 source "${GEN3_HOME}/gen3/lib/utils.sh"
--- a/gen3/bin/healthcheck.sh
+++ b/gen3/bin/healthcheck.sh
@@ -70,7 +70,7 @@ gen3_healthcheck() {
   fi
 
   local healthy=true
-  if [[ !$internetAccess || ! -z "$pendingLongPods" || ! -z "$terminatingPods" || ! -z "$unknownPods" || ! -z "$crashLoopPods" || ! -z "$evictedPods" || ! -z "$notReadyNodes" ]]; then
+  if [[ "$internetAccess" = false || ! -z "$pendingLongPods" || ! -z "$terminatingPods" || ! -z "$unknownPods" || ! -z "$crashLoopPods" || ! -z "$evictedPods" || ! -z "$notReadyNodes" ]]; then
     healthy=false
   fi
 
--- a/gen3/bin/healthcheck.sh
+++ b/gen3/bin/healthcheck.sh
@@ -1,0 +1,18 @@
+#
+# TODO: write info here
+#
+
+source "${GEN3_HOME}/gen3/lib/utils.sh"
+gen3_load "gen3/gen3setup"
+
+help() {
+  gen3 help healthcheck
+}
+
+gen3_healthcheck() {
+  local evictedPods=$(kubectl get pods -o json | \
+    jq  '.items[] | select(.status.reason!=null) | select(.status.reason | contains("Evicted"))')
+  local unknownPods=$(kubectl get pods --field-selector status.phase=Unknown -o json | jq '.items[]')
+  local crashLoopPods=$(kubectl get pods --field-selector status.phase=CrashLoopBackOff -o json | jq '.items[]')
+  local pendingPods=$(kubectl get pods --field-selector status.phase=Pending -o json | jq '.items[]')
+}
--- a/gen3/bin/healthcheck.sh
+++ b/gen3/bin/healthcheck.sh
@@ -103,28 +103,32 @@ gen3_healthcheck() {
   # check internet access
   echo "Checking internet access..." 1>&2
   local curlCmd="curl --max-time 15 -s -o /dev/null -I -w "%{http_code}" https://www.google.com"
+  local statusCode=0
   if [[ $HOSTNAME == *"admin"* ]]; then # if in admin vm, run curl in fence pod
-    curlCmd="g3kubectl exec $(gen3 pod fence) -- $curlCmd"
+    statusCode=$(g3kubectl exec $(gen3 pod fence) -- $curlCmd)
+  else # not inside adminvm, curl from here
+    statusCode=$(eval $curlCmd)
   fi
-  local statusCode=$(eval $curlCmd)
   local internetAccess=true
   if [[ $statusCode -lt 200 || $statusCode -ge 400 ]]; then
     internetAccess=false
   fi
+ 
+  # check internet access with explicit proxy
   echo "Checking explicit proxy internet access..." 1>&2
-  (
-    export http_proxy="http://cloud-proxy.internal.io:3128"
-    export https_proxy=$http_proxy
-    local statusCodeExplicit=$(eval $curlCmd)
-    if [[ $statusCodeExplicit -lt 200 || $statusCodeExplicit -ge 400 ]]; then
-      exit 1
-    else
-      exit 0
-    fi
-  )
-  explicitProxyResult=$?
+  local http_proxy="http://cloud-proxy.internal.io:3128"
+  local statusCodeExplicit=0
+  if [[ $HOSTNAME == *"admin"* ]]; then # inside adminvm, curl from fence pod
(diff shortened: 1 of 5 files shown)


gen3/bin/healthcheck.sh
```diff
@@ -0,0 +1,118 @@
...
+      ;;
+    *)
+      echo "Unrecognized flag"
+      help
+      exit 1
+      ;;
+  esac
+done
+
+gen3_healthcheck
```
frickjack: You can pass command line arguments through to gen3_healthcheck:
    `gen3_healthcheck "$@"`
, then move the argument checking into `gen3_healthcheck` rather than setting the global `HEALTH_SEND_SLACK` variable

gen3/bin/healthcheck.sh
```diff
@@ -0,0 +1,118 @@
...
+  local allNodes=$(g3kubectl get nodes -o json | \
+    jq -r '[
+      .items[] | {
+        name: .metadata.labels."kubernetes.io/hostname", conditions: [ .status.conditions[] | { (.type): .status }] | add
+      }
+    ]')
+  local notReadyNodes=$(echo $allNodes | jq -r '.[] | select(.conditions.Ready != "True")')
+
+  # check internet access
+  local statusCode=$(curl -s -o /dev/null -I -w "%{http_code}" https://www.google.com)
```
frickjack: I'm not sure how curl behaves when the squid is down - in that case curl will fail to even connect to a server, so not statusCode will come back.  You can probably test by disabling your WiFi.

gen3/bin/healthcheck.sh
```diff
@@ -0,0 +1,118 @@
...
+  local internetAccess=true
```
frickjack: Add a second statusCode check with an explicity proxy set to `cloud-proxy.internal.io` - if you set that in a `()` subshell, then it goes away when the subshell completes - ex:
```
(
  export http_proxy="http://cloud-proxy.internal.io:3128"
  export https_proxy=$http_proxy
  statusCode=...
  exit 0 or exit 1 depending on test success
)
explicitProxyResult=$?
```

frickjack: One more thing - it would be cool if you could determine whether the healthcheck is running on the `admin vm` (user just typing `gen3 healthcheck`) or in a pod (cronjob, whatever) on the cluster - we ideally want to test `curl` internet acccess on the cluster - can maybe `kubectl exec ...` into fence, or trick out `gen3 devterm` somehow, or just run a `healthcheck-job.yaml`, and wait for the job to finish, so the healthcheck runs on the cluster ...

gen3/bin/healthcheck.sh
```diff
@@ -0,0 +1,139 @@
...
+  local allNodes=$(g3kubectl get nodes -o json | \
+    jq -r '[
+      .items[] | {
+        name: .metadata.labels."kubernetes.io/hostname", conditions: [ .status.conditions[] | { (.type): .status }] | add
+      }
+    ]')
+  local notReadyNodes=$(echo $allNodes | jq -r '[.[] | select(.conditions.Ready != "True")]')
+
+  # check internet access
+  local curlCmd="curl -s -o /dev/null -I -w "%{http_code}" https://www.google.com"
```
fauzigo: Perhaps you may want to use `--connect-timeout` or `-m` or `--max-time`, you don't want curl to hang until a default timeout which may take a few mins. If after 45 seconds there is not response, definitely down, I would say 15, seconds, but 45 should be the max.

gen3/bin/healthcheck.sh
```diff
@@ -0,0 +1,139 @@
...
+  if [[ "$HEALTH_SEND_SLACK" = true && "$healthy" = false ]]; then
+    if [[ "${slackWebHook}" == 'None' || -z "${slackWebHook}" ]]; then
+      slackWebHook=$(g3kubectl get configmap global -o jsonpath={.data.slack_webhook})
+    fi
+    if [[ "${slackWebHook}" == 'None' || -z "${slackWebHook}" ]]; then
+      echo "WARNING: slackWebHook is None or doesn't exist; not sending results to Slack"
+    else
+      local tmpHostname=$(g3kubectl get configmap manifest-global -o jsonpath={.data.hostname})
+      local formattedAttachment="{\"title\": \"Status\", \"text\": \"$(echo $healthJson | sed s/\"/\\\\\"/g | sed s/,/,\\n/g)\", \"color\": \"#FF0000\"}"
+      curl -X POST --data-urlencode "payload={\"text\": \"Healthcheck FAILED for ${tmpHostname}\n\", \"attachments\": [$formattedAttachment]}" "${slackWebHook}"
```
fauzigo: If curling to the outside to the outside to check for internet access, this one will most likely fail as well, so maybe a timeout flag would also be useful here

gen3/bin/healthcheck.sh
```diff
@@ -0,0 +1,139 @@
...
+	echo "Unrecognized flag"
+	help
+	exit 1
+	;;
+    esac
+  done
+
+  # refer to k8s api docs for pod status info
+  # https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.13/#podstatus-v1-core
+  local allPods=$(g3kubectl get pods -o json | \
```
frickjack: Pass `--all-namespaces` (I think that's it), so we get everything on the cluster (jupyterhub namespaces, kube-system, etc)

gen3/bin/healthcheck.sh
```diff
@@ -0,0 +1,163 @@
...
+  if [[ $HOSTNAME == *"admin"* ]]; then # if in admin vm, run curl in fence pod
+    curlCmd="g3kubectl exec $(get_pod fence) -- $curlCmd"
+  fi
+  local statusCode=$(eval $curlCmd)
+  local internetAccess=true
+  if [[ $statusCode -lt 200 || $statusCode -ge 400 ]]; then
+    internetAccess=false
+  fi
+  (
+    export http_proxy="http://cloud-proxy.internal.io:3128"
```
frickjack: this is tricky, but if we're running on the `adminvm` then we want to do something like:
```
g3kubectl exec $(gen3 pod fence) env https_proxy=... curl https://www.google.com
```

gen3/bin/healthcheck.sh
```diff
@@ -0,0 +1,163 @@
...
+      .items[] | {
+        name: .metadata.labels."kubernetes.io/hostname", conditions: [ .status.conditions[] | { (.type): .status }] | add
+      }
+    ]')
+  local notReadyNodes=$(echo $allNodes | jq -r '[.[] | select(.conditions.Ready != "True")]')
+
+  # check internet access
+  local curlCmd="curl --max-time 15 -s -o /dev/null -I -w "%{http_code}" https://www.google.com"
+  if [[ $HOSTNAME == *"admin"* ]]; then # if in admin vm, run curl in fence pod
+    curlCmd="g3kubectl exec $(get_pod fence) -- $curlCmd"
```
frickjack: Use `gen3 pod fence` rather than `get_pod ...` - might want to get rid of that some day ...

gen3/bin/healthcheck.sh
```diff
@@ -0,0 +1,163 @@
...
+      # check how long it's been terminating
+      local s=$(echo $statusLine | grep -oP "\d+(?=s)")
+      local m=$(echo $statusLine | grep -oP "\d+(?=m)")
+      local h=$(echo $statusLine | grep -oP "\d+(?=h)")
+      if [[ $m -gt 10 || $s -gt 300 || $h -gt 0 ]]; then
+        terminatingTimeoutPods=$(echo $terminatingTimeoutPods | jq -r ". += [$pod]")
+      fi
+    fi
+  done <<< "$(echo $allPods | jq -c '.[]')"
+
```
frickjack: Why are you scanning all pods here?

kube/services/jobs/healthcheck-cronjob.yaml
```diff
@@ -0,0 +1,45 @@
...
+                  value: ""
+                - name: GEN3_NOPROXY
+                  value: none
+              command: ["/bin/bash" ]
+              args:
+                - "-c"
+                - |
+                  set -i
+                  source ~/.bashrc
+                  export JENKINS_HOME=none
```
frickjack: do you need to set JENKINS_HOME?

PR #709, opened 2019-02-27
Feat/gen3 s3

macintoshpie:
### New Features
add basic s3 operations to gen3

### DIFF
--- a/gen3/bin/s3.sh
+++ b/gen3/bin/s3.sh
@@ -61,21 +61,6 @@ EOF
     exit 1
   fi
   gen3 trash
-
-  gen3_log_info "Attempting to add bucket to cloudtrail"
-  local cloudtrailName="${environmentName}-data-bucket-trail"
-  local cloudtrailEventSelectors=$(gen3_aws_run aws cloudtrail get-event-selectors --trail-name $cloudtrailName | jq -r '.EventSelectors')
-  if [[ -z "$cloudtrailEventSelectors" ]]; then
-    # uh oh... for some reason the cloudtrail is not what we expected it to be
-    gen3_log_info "Unable to find cloudtrail with name $cloudtrailName"
-    exit 0
-  fi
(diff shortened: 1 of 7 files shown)


gen3/bin/s3.sh
```diff
@@ -0,0 +1,76 @@
+#!/bin/bash
+#
+# Describe and create s3 buckets
+#
+
+source "${GEN3_HOME}/gen3/lib/utils.sh"
+gen3_load "gen3/gen3setup"
+
+help() {
```
frickjack: name all these function `gen3_s3_...`, so we could potentially `source` this file in the future, and names won't collide ...

gen3/bin/s3.sh
```diff
@@ -0,0 +1,76 @@
...
+    cat << EOF
+ERROR: Bucket name does not meet the following requirements:
+  - starts with a-z
+  - contains only a-z, 0-9, and dashes, "-"
+EOF
+    exit 1
+  fi
+  # if bucket already exists do nothing and exit
+  if [[ -z "$(gen3_aws_run aws s3api head-bucket --bucket $bucketName 2>&1)" ]]; then
+    echo "INFO: Bucket already exists"
```
frickjack: use `gen3_log_err` instead of `echo` - it takes care of piping to `stderr`, and also colors thing red

gen3/bin/s3.sh
```diff
@@ -0,0 +1,76 @@
...
+    gen3_aws_run aws s3 ls s3://$1
+  fi
+}
+
+new() {
+  local bucketName=$1
+  # do simple validation of bucket name
+  local regexp="^[a-z][a-z0-9\-]*$"
+  if [[ ! $bucketName =~ $regexp ]];then
+    cat << EOF
```
frickjack: send info messages to stderr as a general rule - I think this will work:
```
cat 1>&2 <<EOM
...
```

PR #712, opened 2019-02-28
add(TCP traffic): for DNS

fauzigo:

### Bug Fixes
- This fixed the name resolution issue for pods running on the same worker as the kube-dns service. Forcing TCP over UDP in /etc/resolv.conf


fauzigo:
> Let’s add that to usersync cronjob too

Added

PR #713, opened 2019-03-01
chore(updateAndRebootCron): weekly admin vm reboot

frickjack:
update and reboot the admin vm weekly

### DIFF
--- a/files/scripts/updateAndRebootCron
+++ b/files/scripts/updateAndRebootCron
@@ -4,8 +4,8 @@
 #  sudo cp updateAndRebootCron /etc/cron.d/updateAndRebootCron
 #  sudo chmod 0644 /etc/cron.d/updateAndRebootCron
 #
-# Some weirdness about Debian cron.weekly:
-# * file must be owned by root, at most 0755
+# gen3 kube-setup-workvm installs this too
 #
+PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
 
-1 1 * * 0 root (date && apt update && apt upgrade -y && systemctl reboot) > /var/log/updateAndRebootCron.log 2>&1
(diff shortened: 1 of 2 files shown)


files/scripts/updateAndRebootCron
```diff
@@ -0,0 +1,11 @@
...
+#
+# Run this as a cron job:
+#  sudo cp updateAndRebootCron /etc/cron.d/updateAndRebootCron
+#  sudo chmod 0644 /etc/cron.d/updateAndRebootCron
+#
+# Some weirdness about Debian cron.weekly:
+# * file must be owned by root, at most 0755
+#
+
+1 1 * * 0 root (date && apt update && apt upgrade -y && systemctl reboot) > /var/log/updateAndRebootCron.log 2>&1
```
fauzigo: I think you may want to force a non interactive apt run, sometimes it might want to update grub and if running on console through cron would fail, something like

`~/planX/cloud-automation/ grep noninteractive tf_files/aws/modules/utility-vm/cloud.tf
sudo DEBIAN_FRONTEND='noninteractive' apt-get -y -o Dpkg::Options::='--force-confdef' -o Dpkg::Options::='--force-confold' upgrade| sudo tee --append /var/log/bootstrapping_script.log`

gen3/bin/kube-setup-workvm.sh
```diff
@@ -106,6 +106,9 @@ if sudo -n true > /dev/null 2>&1 && [[ $(uname -s) == "Linux" ]]; then
     tar xf "${XDG_RUNTIME_DIR}/helm.tar.gz" -C ${XDG_RUNTIME_DIR}
     sudo mv "${XDG_RUNTIME_DIR}/linux-amd64/helm" /usr/local/bin
   fi
+
+  # install "update and reboot" cron job
+  sudo cp "${GEN3_HOME}/files/scripts/updateAndRebootCron" /etc/cron.d/
```
fauzigo: Shouldn't we check if this already exist before copying into the cron dir ? I know we sometimes run workvm more than once.

frickjack: I was thinking we would overwrite the file even if it already exists, so we always deploy the latest file ....

PR #717, opened 2019-03-06
Fix/reset

frickjack:
couple small tweaks:
* aws token timeout
* no `set -e` (from kube-setup-init) in `gen3 reset`

### DIFF
--- a/gen3/lib/aws.sh
+++ b/gen3/lib/aws.sh
@@ -34,9 +34,11 @@ gen3_aws_run() {
 
     # Try to use cached creds if possible
     if [[ -f $gen3CredsCache ]]; then
+      local tPlus5Mins
+      tPlus5Mins=$(($(date +%s) + 300))
       gen3AwsExpire=$(jq -r '.Credentials.Expiration' < $gen3CredsCache)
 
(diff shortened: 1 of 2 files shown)


gen3/lib/aws.sh
```diff
@@ -34,9 +34,11 @@ gen3_aws_run() {
 
     # Try to use cached creds if possible
     if [[ -f $gen3CredsCache ]]; then
+      local tPlus5Mins
+      tPlus5Mins=$(($(date +%s) + 300))
       gen3AwsExpire=$(jq -r '.Credentials.Expiration' < $gen3CredsCache)
 
-      if [[ "$gen3AwsExpire" =~ ^[0-9]+ && "$gen3AwsExpire" > "$(date -u +%Y-%m-%dT%H:%M)" ]]; then
+      if [[ "$gen3AwsExpire" =~ ^[0-9]+ && "$gen3AwsExpire" > "$(date --utc --date "@$tPlus5Mins" +%Y-%m-%dT%H:%M)" ]]; then
```
fauzigo: Isn't better and easier to do something like `date -d "+5 mins"` instead

PR #720, opened 2019-03-07
add(conditional): checks

fauzigo:

### Bug Fixes
- If your namespace is not default, then the revproxy config won't be added to point to prometheus/grafana

### DIFF
--- a/gen3/bin/kube-setup-revproxy.sh
+++ b/gen3/bin/kube-setup-revproxy.sh
@@ -14,7 +14,8 @@ set -e
 source "${GEN3_HOME}/gen3/lib/utils.sh"
 gen3_load "gen3/gen3setup"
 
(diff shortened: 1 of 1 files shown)


gen3/bin/kube-setup-revproxy.sh
```diff
@@ -14,6 +14,8 @@ set -e
 source "${GEN3_HOME}/gen3/lib/utils.sh"
 gen3_load "gen3/gen3setup"
 
+current_namespace=$(g3kubectl config view -o jsonpath={.contexts[].context.namespace})
+
```
frickjack: this doesn't work in Jenkins
would be good to also check the KUBECTL_NAMESPACE env variable
you can just call:
```
current_namespace=$(gen3 db namespace)
```
, or check what that code does if you don't want to do that ... :-p

PR #747, opened 2019-03-20
fix(jupyter): namespace role binding

philloooo:
This allows doing parallel QA in jupyterhub in our QA environments sharing the same jupyter namespace.
It doesn't allow a test user in both QA envs to use jupyter at the same time

### Improvements
- namespace the rolebinding name for service account used by jupyterhub in gen3 services' namespace

### DIFF
--- a/gen3/bin/kube-setup-jupyterhub.sh
+++ b/gen3/bin/kube-setup-jupyterhub.sh
@@ -25,7 +25,8 @@ gen3 update_config jupyterhub-config "${GEN3_HOME}/kube/services/jupyterhub/jupy
 
 g3kubectl apply -f "${GEN3_HOME}/kube/services/jupyterhub/serviceaccount.yaml"
 g3kubectl apply -f "${GEN3_HOME}/kube/services/jupyterhub/role-jupyter.yaml"
-g3kubectl apply -f "${GEN3_HOME}/kube/services/jupyterhub/rolebinding-jupyter.yaml"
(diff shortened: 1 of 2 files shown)


gen3/bin/kube-setup-jupyterhub.sh
```diff
@@ -25,7 +25,8 @@ gen3 update_config jupyterhub-config "${GEN3_HOME}/kube/services/jupyterhub/jupy
 
 g3kubectl apply -f "${GEN3_HOME}/kube/services/jupyterhub/serviceaccount.yaml"
 g3kubectl apply -f "${GEN3_HOME}/kube/services/jupyterhub/role-jupyter.yaml"
-g3kubectl apply -f "${GEN3_HOME}/kube/services/jupyterhub/rolebinding-jupyter.yaml"
+namespace="$(gen3 db namespace)"
+g3k_kv_filter ${GEN3_HOME}/kube/services/jupyterhub/rolebinding-jupyter.yaml JUPYTER_BINDING "name: jupyter-binding-$namespace" CURRENT_NAMESPACE "namespace: $namespace" | g3kubectl apply -f -
```
ZakirG: are you sure `name: jupyter-binding-$namespace` is correct? a hardcoded `name: jupyter-binding` is what works for me

philloooo: seems working in qa dcp and niaid

PR #749, opened 2019-03-20
Feat/manifest service deploy

ZakirG:
Don't merge this PR!!! It will again be altered once Fauzi has updated the Jupyter pod kernels


kube/services/jupyterhub/jupyterhub_config.py
```diff
@@ -94,6 +116,12 @@
         }
     }
 }
+# TODO: Remove this below line. But need this here for now for integration test purposes.
```
frickjack: Did you mean to remove some of this?

ZakirG: Nahh, I'm just submitting for review prematurely. Will remove that later

gen3/bin/kube-setup-manifestservice.sh
```diff
@@ -0,0 +1,18 @@
...
+#
+# Deploy manifestservice into existing commons
+# This is an optional service that's not part of gen3 core services
+# It only needs to be deployed to commons that have Export to Workspace functionality
+
+source "${GEN3_HOME}/gen3/lib/utils.sh"
+gen3_load "gen3/gen3setup"
+
+gen3 kube-setup-secrets
+
```
frickjack: You can add calls to setup the S3 bucket that manifest-services wants too:
```
hostname="$(g3kubectl get configmap global -o json | jq -r .data.hostname)"
bucketname="manifest-${hostname//./-}"
gen3 s3 create "$bucketname"
# this will fail if manifest-user does not exist ... I think
gen3 s3 attach-bucket-policy "$bucketname" --read-write --user-name manifest-user || true
```
We still have to creat the `manifest-user` and `creds.json` by hand for now ...

ZakirG: Are s3 buckets cheap? Is it fine if I test this code?

philloooo:
substituted by #786

PR #754, opened 2019-03-22
Feat/custom kernel

fauzigo:

### New Features
- The ability of making jupyter nodepool's nodes come up with a different kernel. 
- Bootstrapscripts can be chosen for each nodepool.


flavors/eks/bootstrap-3.0.0.sh
```diff
@@ -0,0 +1,59 @@
...
+# Using `aws s3 sync` throws all kind of issues in the logs, using cp instead. 
+
+aws s3 cp --recursive $S3_LOCATION $KERNEL_FILES
+
+rpm -iUv $KERNEL_FILES/kernel*.rpm
+
+
+
+# When the host is back up after reboot, it should trigger the eks/bootstrap script for it to join th cluster.
+chmod +x /etc/rc.d/rc.local
```
frickjack: can we do this with `systemd`?  It seems goofy to use systemd for some stuff, then generate old-school init bash scripts for other things ...

PR #775, opened 2019-04-11
fix(guppy): remove unused env var, and allow large size queries

qingyashu:
### Deployment changes
 - Remove unused env var from guppy config, and allow large size queries


kube/services/revproxy/gen3.nginx.conf/guppy-service.conf
```diff
@@ -3,4 +3,5 @@
               set $upstream_guppy http://guppy-service.$namespace.svc.cluster.local;
               rewrite ^/guppy/(.*) /$1 break;
               proxy_pass $upstream_guppy;
+              client_max_body_size 0;
```
m0nhawk: Is it really necessary to disable client request body size?

qingyashu: It's for asking manifest entries (e.g. get `file_id` with a lot of `subject_id`s in filters), the request payload will be big

PR #777, opened 2019-04-12
feat(create-fuse-sidecar): Updated kubespawner in jupyterhub to spawn…

emalinowski:
… sidecar container

### New Features
- Added sidecar container to jupyterhub which will grab the most updated manifest and fuse mount it to the shared /data filesystem.

### DIFF
--- a/kube/services/jupyterhub/jupyterhub_config.py
+++ b/kube/services/jupyterhub/jupyterhub_config.py
@@ -118,11 +118,8 @@ c.KubeSpawner.lifecycle_hooks = {
     }
 }
 # TODO: Remove this below line before merge to master. But need this here for now for testing purposes.
-c.KubeSpawner.image_pull_policy = 'Always'
+#c.KubeSpawner.image_pull_policy = 'Always'
 c.KubeSpawner.modify_pod_hook = modify_pod_hook
-c.KubeSpawner.cmd = 'start-singleuser.sh'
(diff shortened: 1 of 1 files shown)


kube/services/jupyterhub/jupyterhub_config.py
```diff
@@ -107,3 +136,15 @@
 c.JupyterHub.hub_ip = "0.0.0.0"
 c.RemoteUserAuthenticator.auth_refresh_age = 1
 c.RemoteUserAuthenticator.refresh_pre_spawn = True
+
+c.KubeSpawner.env_keep = ['HOSTNAME', 'POD_NAMESPACE']
+
+c.KubeSpawner.extra_containers = [{
```
frickjack: add a `fuseEnabled` flag to the `jupyter` block of `manifest.json` (you can pass that into the jupyterhub container as an environment variable by modifying the jupyterhub-deploy.yaml), and only run the block that adds the fuse-sidecar if `fuseEnabled` is set true - ex:
```
if os.environ.get('fuseEnabled') == 'yes':
    c.KubeSpawner.extra_containers = [ ... ]
```

philloooo: I don't think we have jupyter that doesn't need fuse sidecar. So I don't see a flag is needed..

PR #780, opened 2019-04-12
fix(gen3_s3): echo result of bucket exists

macintoshpie:
### Bug Fixes
- was previously attempting to return a string as the result for checking if a bucket existed... now echos result


macintoshpie:
Closing as this was already fixed by #786

PR #793, opened 2019-04-19
feat(arborist): setup arborist db and new deployment

Avantol13:
### New Features
- arborist db creation using gen3 db command, added new arborist deployment

### Deployment changes
- new arborist requires a database, if you update arborist version in the manifest, must also specify a new "arborist" block in your manifest with `deployment_version` == `2`


kube/services/arborist/arborist-deploy-2.yaml
```diff
@@ -0,0 +1,76 @@
...
+        - name: "arborist-secret"
+          readOnly: true
+          mountPath: "/var/www/arborist/dbcreds.json"
+          subPath: dbcreds.json
+        command: ["sh"]
+        args:
+          - "-c"
+          - |
+            # set env vars
+            export PGDATABASE=$(cat /var/www/arborist/dbcreds.json | jq -r '.db_database')
```
rudyardrichter: there seems to be some weirdness with putting environment variables here, I'm not sure if they stick around in the pod:
```bash
rudyard@cdistest_dev_admin ~
$ g3kubectl exec -it $(gen3 pod arborist) -- /bin/bash
bash-4.4# export | grep PG
bash-4.4#
```
which means the server will find the database ok, but you can't run the migrations:
```
rudyard@cdistest_dev_admin ~
$ g3kubectl exec $(gen3 pod arborist) -- /go/src/github.com/uc-cdis/arborist/migrations/latest
current datase version:        0000-00-00T000000Z
applying migration:            2019-02-18T214320Z_init
problem during last migration step, exiting early:
psql: could not connect to server: No such file or directory Is the server running locally and accepting connections on Unix domain socket "/tmp/.s.PGSQL.5432"?
command terminated with exit code 1
```

Avantol13: that's really strange... b/c the endpoints seem to function (adding stuff to db, deleting etc). You're right though, I don't see the env vars when I exec in...

PR #798, opened 2019-04-24
PXD-2679 Feat/refresh service

giangbui:

### New Features
- Generate refresh data reports, augmented manifests and do validation


gen3/bin/dcf.sh
```diff
@@ -0,0 +1,183 @@
...
+  if [ -z "$manifest" ]; then
+    echo "fail to download $manifest_type manifest!!!"
+    exit 1
+  fi
+
+  refresh_log=${manifest//.tsv/.txt}
+
+  echo "$refresh_log"
+
+  if [ ! -f "/tmp/$manifest" ]; then
```
frickjack: Use `mktemp` to setup temporary files, and avoid using `/tmp/` - $XDG_RUNTIME_DIR points to a temp directory that is particular to the logged in user - ex:
```
manifestLocal="$(mktemp $XDG_RUNTIME_DIR/manifest_XXXXXX)"
```

gen3/bin/dcf.sh
```diff
@@ -0,0 +1,183 @@
...
+AWS_OUTPUT_BUCKET="data-refresh-output"
+GS_INPUT_BUCKET="replication-input"
+GS_OUTPUT_BUCKET="datarefresh-log"
+
+echo "Hello from dcf!"
+command=$1
+release=$2
+
+# Activate google service account
+if [ -f "${WORKSPACE}/${vpc_name}/apis_configs/dcf_dataservice/creds.json" ]; then
```
frickjack: Use the `g3auto` folder for secrets.  Also - if it's not too much trouble - use a different name than `creds.json`, since we already use that for database creds for other services - maybe `googleCreds.json` - something like that ex:
```
$(gen3_secrets_folder)/g3auto/dcf_dataservice/googleCreds.json
```

gen3/bin/dcf.sh
```diff
@@ -0,0 +1,183 @@
...
+  echo "Validate aws refresh"
+
+  manifest_type=$3
+  manifest="$(aws s3 ls s3://$AWS_INPUT_BUCKET | grep GDC_full_sync_${manifest_type}_.*$release | awk -F' ' '{print $4}')"
+  
+  if [ -z "$manifest" ]; then
+    echo "fail to download $manifest_type manifest!!!"
+    exit 1
+  fi
+
```
frickjack: See the earlier comment about using /tmp/.  If what you really want to do is cache a file locally, then use `$GEN3_CACHE_DIR` instead, but it's not a big deal to just copy the file locally for a command line tool to avoid synchronization issues.

gen3/bin/dcf.sh
```diff
@@ -0,0 +1,183 @@
...
+  
+  validation_log="$(aws s3 ls s3://$AWS_OUTPUT_BUCKET/$release/ | grep validation.log | awk -F' ' '{print $4}')"
+
+  if [ -z "${validation_log}" ]; then
+    echo "The validation log does not exist. Please run the validation script first
+    as desribed in https://github.com/uc-cdis/cdis-wiki/blob/master/ops/Data-refresh.md
+    """
+    exit 1
+  fi
+
```
frickjack:
```
localFile="$(mktemp "$XDG_RUNTIME_DIR/validation_XXXXXX")"
```

gen3/bin/dcf.sh
giangbui: I made the change in d94ef65a821d039b329d3e21533b9a9d809fef7c
Thanks!

gen3/bin/dcf.sh
giangbui: I cache the file here so that we don't have to re-download the manifest files again. Thanks!

PR #822, opened 2019-05-09
feat/grafana-dashboards

m0nhawk:


m0nhawk:
> * I think you need to reference the new dashboard in `grafana-values.yaml`
Grafana doesn't require listing the provisioning dashboards.

> * Can we not use spaces in file names?
Fixed. Grafana by default saving with the actual dashboard name.

> * README would be nice
Added.

PR #839, opened 2019-05-15
Patch/gitops sync portal

macintoshpie:
### Changes
- gitops-sync now uses locks to prevent stepping on each others toes when rolling

### Bug Fixes
- fixed detecting diffs in portal configuration for gitops sync

### DIFF
--- a/gen3/bin/gitops.sh
+++ b/gen3/bin/gitops.sh
@@ -56,6 +56,7 @@ gen3_gitops_sync() {
   local dictAttachment
   local versionsAttachment
   local commonsManifestDir
+  local portalDiffs
 
   if [[ $1 = '--slack' ]]; then
     if [[ "${slackWebHook}" == 'None' || -z "${slackWebHook}" ]]; then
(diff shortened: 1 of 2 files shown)


gen3/bin/gitops.sh
```diff
@@ -123,35 +124,63 @@ gen3_gitops_sync() {
...
-      elif [[ "${oldFile}" != "${newFile}" ]]; then
-        gen3_log_info "Found diff in portal/${filename} - difference between file and secret"
+    local secretsFile
+    local comparingFile
+    local diffMsg
+    for defaultFilepath in $(find "${defaultsDir}/" -name "gitops*" -type f); do
+      diffMsg=""
+      filename=$(basename ${defaultFilepath})
+      commonsFilepath="${commonsManifestDir}/portal/$filename"
+      secretsFile=$(g3kubectl get secret portal-config -o json | jq -r '.data."'"${filename}"'"')
```
frickjack: do `.data["$filename"]` - data.bla breaks if bla has unexpected characters in it

macintoshpie: I have it wrapped in " ", which does the same thing as [ ]

PR #851, opened 2019-05-20
feat(indexd): pull arborist config from manifest

Avantol13:
### Improvements
- Indexd deployment will contain arborist service if manifest entry specifies indexd should run with arborist

### DIFF
--- a/apis_configs/indexd_settings.py
+++ b/apis_configs/indexd_settings.py
@@ -28,7 +28,7 @@ dist = environ.get("DIST", None)
 if dist:
     CONFIG["DIST"] = json.loads(dist)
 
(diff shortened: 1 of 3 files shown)


apis_configs/indexd_settings.py
```diff
@@ -28,6 +28,8 @@ def load_json(file_name):
 if dist:
     CONFIG["DIST"] = json.loads(dist)
 
+arborist = environ.get("ARBORIST", False)
```
frickjack: Maybe do something like:
```
arborist = (environ.get("ARBORIST", "False") == "True")
```
, otherwise if someone puts "False" or "True" in the manifest - this check will evaluate `True` below ...

PR #864, opened 2019-05-29
chore(authz-migration): add job to set authz fields

rudyardrichter:
### New Features

- Add `indexd-authz` job to set the `authz` fields in index records according to their existing ACLs.


kube/services/jobs/indexd-authz-job.yaml
```diff
@@ -0,0 +1,69 @@
+apiVersion: batch/v1
```
Avantol13: talked about this in person but want to note here: we should add arguments for passing through start did and whether or not to use sheepdog db for checking program vs. project

PR #870, opened 2019-06-05
fix(reset): Jenkins complexity with wts

frickjack:
Fix handling of wts secret by `gen3 reset`

### DIFF
--- a/gen3/bin/reset.sh
+++ b/gen3/bin/reset.sh
@@ -172,7 +172,7 @@ g3kubectl create configmap fence "--from-file=user.yaml=$useryaml"
 run_setup_jobs
 clear_wts_clientId
 gen3 roll all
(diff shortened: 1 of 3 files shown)


gen3/bin/reset.sh
```diff
@@ -145,8 +170,10 @@ g3kubectl create configmap fence "--from-file=user.yaml=$useryaml"
 # try to make reset more reliable - especially in Jenkins
 #
 run_setup_jobs
-remove_wts_creds_secrets
+clear_wts_clientId
 gen3 roll all
+reset_wts_clientId
```
mfshao: there is no `reset_wts_clientId`, do you mean `new_wts_clientId` ?

frickjack: doh!

PR #896, opened 2019-06-19
chore(kube-setup-metrics): initial

frickjack:
setup k8s metrics server - prerequisite for horizontal auto scaling


tf_files/aws/modules/admin-vm/cloud.tf
```diff
@@ -161,10 +161,13 @@ resource "aws_instance" "login" {
   monitoring             = true
   key_name               = "${var.ssh_key_name}"
   vpc_security_group_ids = ["${aws_security_group.ssh.id}", "${aws_security_group.local.id}"]
-
   #vpc_security_group_ids = ["${var.secgroup_adminvms}"]
   iam_instance_profile = "${aws_iam_instance_profile.child_role_profile.name}"
 
+  root_block_device  {
+    volume_size = 8
```
fauzigo: seems like 8 to me. Maybe a better idea to make it a variable with 24 as default value just in case?

PR #908, opened 2019-07-05
feat/privacy policy: support for fence & windmill serving privacy policy

rudyardrichter:
TODO:
- [x] fence changes
- [x] windmill changes

### New Features

- add support for fence serving privacy policy
    - add `PRIVACY_POLICY_URL` env for fence which defines redirect
    - add volume mount for static `.md` file to put the policy in

In terms of deployment changes, be sure to run `kube-setup-secrets` before using these changes, otherwise the fence and data-portal pods will fail to come up because of the missing volume.

### DIFF
--- a/kube/services/fence/fence-deploy.yaml
+++ b/kube/services/fence/fence-deploy.yaml
@@ -82,6 +82,9 @@ spec:
         - name: ca-volume
           secret:
             secretName: "service-ca"
+        - name: privacy-policy
+          configMap:
+            name: "privacy-policy"
       containers:
       - name: fence
         GEN3_FENCE_IMAGE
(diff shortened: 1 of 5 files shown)


kube/services/fence/fence-deploy.yaml
```diff
@@ -90,6 +93,12 @@ spec:
           value: /var/www/fence
         - name: GEN3_DEBUG
           GEN3_DEBUG_FLAG|-value: "False"-|
+        - name: PRIVACY_POLICY_URL
```
Avantol13: there are other fence deployment yamls you may need to update with this information. you can just grep `name: fence` to see where fence containers are used, noteably, you probably want to update the `fence-canary-deploy`. Unsure if `fenceshib-deploy` really needs this.. there are also a bunch of other jobs that include fence but most of those just call out to fence-create. So long as fence behaves when this config is not there then that should be fine

rudyardrichter: Yeah, probably good to add to the canary (and maybe fenceshib?). That's correct, it isn't necessary for pods running fence for jobs like usersync

rudyardrichter: think I will leave out fenceshib for now since this feature won't be available on it until backport/update anyways

PR #955, opened 2019-08-13
Chore/improve k8s report

fauzigo:

### Improvements
- A little more complete report


gen3/bin/report-tool.sh
```diff
@@ -0,0 +1,139 @@
...
+  echo
+  echo -e "AMIs: "${images_id}
+
+}
+
+## Let's do some admin work to find out the variables to be used here
+BOLD='\e[1;31m'         # Bold Red
+REV='\e[1;32m'       # Bold Green
+
+#Help function
```
frickjack: Put a `.md` file under the `doc/` folder so `gen3 help ...` works

gen3/bin/report-tool.sh
```diff
@@ -0,0 +1,139 @@
...
+
+
+function main() {
+
+  case "${1}" in
+    "full")
+      kube_facts
+      instances_facts
+      print_report_full
+      ;;
```
frickjack: not crazy about this pattern of setting global variables for use in other functions.
would be nicer if each function just used local variables, and maybe returned a json blob:
```
kube_facts() {
  ...
  cat - <<EOM
{
    "version": "$version",
   ...
}
EOM
}

...
local k8sInfo=$(kube_facts)
echo "k8s version: $(jq -r .version <<<"$k8sInfo")"
```

we can let it slide this time ... :-p

PR #974, opened 2019-09-06
fix(dev-namespace): tweaks - avoid parent secrets

frickjack:
small tweaks - not able to reproduce issue with secrets corruption in parent namespae


gen3/bin/kube-dev-namespace.sh
```diff
@@ -92,7 +88,10 @@ mkdir -p credentials
 for name in ca.pem ca-key.pem; do
   cp ~/${vpc_name}/credentials/$name credentials/
 done
-sed -i.bak "s/default/$namespace/" kubeconfig
+
+cp kubeconfig kubeconfig.bak
+rm kubeconfig
```
fauzigo: If you are overwriting the file later, there is no need to delete it, but no big deal.

PR #987, opened 2019-09-18
Chore/rds module

fauzigo:

### New Features
- RDS module, to create RDS instances using `__rds`


tf_files/aws/modules/rds/variables.tf
```diff
@@ -0,0 +1,278 @@
...
+  description = "The name of the RDS instance, if omitted, Terraform will assign a random, unique identifier"
+  type        = "string"
+}
+
+variable "rds_instance_username" {
+  description = "Username to use"
+#  default     = ""
+}
+
+variable "rds_instance_password" {
```
frickjack: Can we use this instead of a variable?  https://www.terraform.io/docs/providers/random/r/password.html

Also take a look at this: https://www.terraform.io/docs/providers/aws/r/secretsmanager_secret_version.html

tf_files/aws/modules/rds/main.tf
```diff
@@ -0,0 +1,168 @@
...
+  count = "${var.rds_instance_create_monitoring_role ? 1 : 0}"
+
+  role       = "${aws_iam_role.enhanced_monitoring.name}"
+  policy_arn = "arn:aws:iam::aws:policy/service-role/AmazonRDSEnhancedMonitoringRole"
+}
+
+resource "aws_db_instance" "this" {
+  count = "${var.rds_instance_create && false == local.is_mssql ? 1 : 0}"
+
+  identifier                            = "${var.rds_instance_identifier}"
```
frickjack: If it's a postgres database - can we do the trick to crank up the limit on the number of connections?

fauzigo: that kind of configurations are managed by "Parameter groups" which is a different resource. We could create a new module that would create that kind of resource and use the output as input for this one. 

Adding it in here would defeat the purpose of simplistic modules.

tf_files/aws/modules/rds/variables.tf
fauzigo: Random added

PR #1001, opened 2019-10-01
squid whitelist += login.bionimbus.org

paulineribeyre:
Add login.bionimbus.org/Shibboleth.sso/DiscoFeed to squid whitelist to fetch list of available shibboleth IDPs from data-portal for incommon

### DIFF
--- a/files/squid_whitelist/web_whitelist
+++ b/files/squid_whitelist/web_whitelist
@@ -66,6 +66,7 @@ k8s.gcr.io
 keyservice.opensciencedatacloud.org
 ks.osdc.io
 lib.stat.cmu.edu
+login.bionimbus.org/Shibboleth.sso/DiscoFeed
 login.mathworks.com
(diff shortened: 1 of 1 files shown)


files/squid_whitelist/web_whitelist
```diff
@@ -66,6 +66,7 @@ k8s.gcr.io
 keyservice.opensciencedatacloud.org
 ks.osdc.io
 lib.stat.cmu.edu
+login.bionimbus.org/Shibboleth.sso/DiscoFeed
```
fauzigo: Squid whitelists domain and IP addresses, there is no need to use URL, won't work.

Correct and push before merging please

paulineribeyre: so will adding just `login.bionimbus.org` work?

fauzigo: yes

PR #1004, opened 2019-10-04
Get ARBORIST_URL from manifest

paulineribeyre:
### Deployment changes
- Sheepdog and Peregrine can be deployed with a custom ARBORIST_URL


apis_configs/peregrine_settings.py
```diff
@@ -13,8 +13,6 @@ def load_json(file_name):
 config["AUTH_ADMIN_CREDS"] = None
 config["INTERNAL_AUTH"] = None
 
-config["ARBORIST"] = "http://arborist-service/"
-
 # Signpost: deprecated, replaced by index client.
```
frickjack: you better leave this alone for old versions of sheepdog and peregrine

PR #1005, opened 2019-10-04
Feat/service account iam

fauzigo:

### New Features
- IAM/service account for kubernetes


gen3/bin/iam-serviceaccount.sh
```diff
@@ -0,0 +1,604 @@
...
+# For more information visit the folowing link
+# https://aws.amazon.com/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/
+
+source "${GEN3_HOME}/gen3/lib/utils.sh"
+gen3_load "gen3/lib/kube-setup-init"
+
+SCRIPT=$(basename ${BASH_SOURCE[0]})
+ACOUNT_ID=$(aws sts get-caller-identity --query Account --output text)
+
+LOGS_SCRIPT_FILE="/tmp/iam-serviceaccount.log"
```
frickjack: Could be weird with multiple users.  Maybe do something like:
```
LOGS_SCRIPT_FILE="$(mktemp "$XDG_RUNTIME_DIR/iam-serviceaccount.log_XXXXXX")"
```

frickjack: you can also use `gen3_log_info` and `gen3_log_err` to log messages to STDERR rather than send them to a log file - ex:
```
gen3_log_info "done doing whatever"
...
gen3_log_err "whatever failed"
...
```

gen3/bin/iam-serviceaccount.sh
```diff
@@ -0,0 +1,604 @@
...
+##
+#
+# function to just print and error and exit 
+#
+# @arg string message to print
+#
+##
+function print_error_and_exit() {
+  local message="${1}"
+  exit 2
```
frickjack: no print?

tf_files/aws/modules/eks/cloud.tf
```diff
@@ -718,3 +718,60 @@ resource "null_resource" "config_setup" {
     "aws_autoscaling_group.eks_autoscaling_group",
   ]
 }
+
+
+
+#--------------------------------------------------------------
+# let's work towards EKS IAM-ServiceAccount integration
+
+resource "aws_iam_openid_connect_provider" "identity_provider" {
```
frickjack: Very cool!  Do we need to add some kind of check to remind the user to update EKS to 1.13+ as part of this?

fauzigo: I couldn't find and elegant way to do it, but I'll try something before merging.

gen3/bin/iam-serviceaccount.sh
fauzigo: Going to take this function out, I thought of replacing `echo`s around with the function, but seemed like more work, and not very practical, will maybe find a better way to do it.

gen3/bin/iam-serviceaccount.sh
fauzigo: OK, sound like a good idea.

tf_files/aws/modules/eks/cloud.tf
fauzigo: 1.11 is coming EOL soon, and 1.12 might also go away in the near future, instead of adding controls that won't make any difference soon we shouldn't spend more time on this.

PR #1006, opened 2019-10-07
feat(authz-mapping): expose authz mapping endpoint

vpsx:
related Arborist PR here https://github.com/uc-cdis/arborist/pull/110 

### New Features
expose arborist's /auth/mapping endpoint (to GET requests only, and without passing the username query param)

### DIFF
--- a/kube/services/revproxy/gen3.nginx.conf/arborist-service.conf
+++ b/kube/services/revproxy/gen3.nginx.conf/arborist-service.conf
@@ -56,6 +56,7 @@ location = /authz/mapping {
         return 403 "failed csrf check";
     }
 
+    # Do not expose POST /auth/mapping
     limit_except GET {
         deny all;
     }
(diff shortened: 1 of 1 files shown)


kube/services/revproxy/gen3.nginx.conf/arborist-service.conf
```diff
@@ -51,6 +51,16 @@ location = /authz/resources {
...
 }
 
+location = /authz/mapping {
+    if ($csrf_check !~ ^ok-\S.+$) {
+        return 403 "failed csrf check";
+    }
+
+    set $upstream http://${arborist_release_name}-service.$namespace.svc.cluster.local;
+
+    proxy_pass $upstream/auth/mapping?username=$arg_username;
```
frickjack: what is $arg_username?

vpsx: the `username` query string for the endpoint--if I leave it out then nginx seems to not pass the query args to arborist

PR #1009, opened 2019-10-08
add(script): report script

fauzigo:

### New Features
- Active users and certificate expiration dates report for the vpn server


files/openvpn_management_scripts/vpn_users_report_generator.sh
```diff
@@ -0,0 +1,111 @@
+#!/bin/bash 
+
```
frickjack: This is awesome!
Would you put some kind of note that this is run on the VPN servers in a cron job under whatever user and generates a report that lists the users with access to the VPN and when their cert expires and where the report is accessed - just some context for someone who comes across the script?

PR #1033, opened 2019-11-16
chore(no-cheaper): run 2 processes by default

paulineribeyre:
After comparing Fence performance with and without uwsgi cheaper, we know it's better to disable it

### Improvements
- Disable uswgi cheaper mode by default in the Python 3 base dockerfile


Docker/python-nginx/python3.6-alpine3.7/Dockerfile
```diff
@@ -165,11 +165,10 @@ COPY supervisord.ini /etc/supervisor.d/supervisord.ini
...
 ENV UWSGI_INI /app/uwsgi.ini
 
-# By default, run 2 processes
-ENV UWSGI_CHEAPER 2
-
-# By default, when on demand, run up to 16 processes
-ENV UWSGI_PROCESSES 16
+# By default, disable uwsgi cheaper mode and run 2 processes.
+# If UWSGI_CHEAPER is set, UWSGI_CHEAPER is the minimum and UWSGI_PROCESSES is the maximum number of processes.
+ENV UWSGI_CHEAPER=
```
Avantol13: shouldn't we just not set this at all instead of setting it to empty?

paulineribeyre: for posterity: we are disabling cheaper mode explicitly - if we need to overwrite the env vars in one of the services we'll know exactly what to do

PR #1055, opened 2019-12-30
chore(load_test):Allocate 0.5 cpu for fence deployment to improve its throughput

themarcelor:
The load testing exercises indicated this change brings an improvement on the service throughput / higher RPS rate, i.e., better performance while working on PreSigned URLs.


emalinowski:
Might have to update the expected value, https://github.com/uc-cdis/cloud-automation/blob/master/gen3/lib/testData/test1.manifest.g3k/expectedFenceResult.yaml#L176.

PR #1067, opened 2020-01-16
Chore/fence config

frickjack:
* optionally pull `fence-config-public.yaml` in from `manifest-fence` configmap - ex:
```
$ cat cdis-manifest/reuben.planx-pla.net/manifests/fence/fence-config-public.yaml
ENABLE_DB_MIGRATION: true

```
Note: default `fence-config-public.yaml` is just an empty file.

* run a `yaml_merge` at launch time in fence deployments and jobs - public config values take precedence:
```
            echo "${FENCE_PUBLIC_CONFIG:-""}" > "/var/www/fence/fence-config-public.yaml"
            python /var/www/fence/yaml_merge.py /var/www/fence/fence-config-public.yaml /var/www/fence/fence-config-secret.yaml > /var/www/fence/fence-config.yaml
```
It would be good to move that logic into fence itself ...


kube/services/jobs/config-fence-job.yaml
```diff
@@ -47,7 +47,7 @@ spec:
           - name: PYTHONPATH
             value: /var/www/fence
           - name: CONVERT_OLD_CFG
-            GEN3_CONVERT_OLD_CFG
+            GEN3_CONVERT_OLD_CFG|-value: true-|
```
Avantol13: we probably want this to be false/empty by default, this is a one-time thing to convert from local_settings.py to the new config, which I think we've already done in all the commons

frickjack: ok :+1:

apis_configs/yaml_merge.py
```diff
@@ -0,0 +1,28 @@
...
+
+import sys
+import json
+from yaml import safe_load as yaml_load
+
+config1 = yaml_load(open(sys.argv[1]))
+config2 = yaml_load(open(sys.argv[2]))
+
+if config1 != None:
+  for key in config1.keys():
```
Avantol13: this will only replace top-level keys, so you can't have "secret" nested keys. there are a lot of instances in the current fence config where nested vars are a mix of public and secret. this is not ideal, but fine as an mvp. The default config handling has the same limitation, if the top-level key is there it assumes the nested vars are too :man_shrugging:

frickjack: yeah - I think it gets confusing if we start dong deep merge.  Hopefully over time we can restructure the fence-config, so that we have distinct public and private keys at the top level of the yaml file ...

kube/services/arborist/arborist-deploy.yaml
```diff
@@ -106,6 +106,15 @@ spec:
...
                 key: sync_from_dbgap
+          - name: FENCE_PUBLIC_CONFIG
+            valueFrom:
+              configMapKeyRef:
+                name: manifest-fence
+                key: fence-config-public.yaml
+                optional: true
+          - name: YAML_MERGE_SCRIPT
+            value: |
+GEN3_YAML_MERGE
```
Avantol13: I assume it was just easier to include the script this way, rather than writing an actual file during setup-secrets and loading into configmaps/secrets and then mounting it in here?

frickjack: yeah - good call - don't know why I talked myself into that :face_with_head_bandage:

Jenkinsfile
```diff
@@ -46,7 +47,7 @@ node {
       }
     }
     stage('SelectNamespace') {
-      (kubectlNamespace, lock) = kubeHelper.selectAndLockNamespace(pipeConfig['UID'])
```
diw2: Sorry just saw this. Andrew made a same change on Jenkinsfile.  https://github.com/uc-cdis/cloud-automation/pull/1068  Post here just in case any conflict shows up.

PR #1080, opened 2020-02-04
chore(packer): copy packer/

frickjack:
retire github.com/uc-cdis/images - move to packer/ subfolder


packer/configs/gituserlist
```diff
@@ -0,0 +1,11 @@
+fauzigo
+philloooo
```
fauzigo: I think we should get out those who are not working for CTDS anymore

frickjack: good catch :+1:

PR #1086, opened 2020-02-11
UWSGI_PROCESSES=1 for fence init pod

mpingram:
- May resolve fence migration startup issue -- if the migration issue is caused by two or more uwsgi processes in the fence-init pod attempting to run migrations simultaneously, as opposed to two or more fence-init pods attempting to run migrations simultaneously.


themarcelor:
Closing this as DevOps will trigger the db migration k8s job prior to the scheduling of the fence pods.

PR #1088, opened 2020-02-12
Python Script to create a ebs volume report and check to see if volumes are encrypted

rbizzell40:
This pull request is to add a python script that will gather ebs volume report into a csv file and upload it results to an s3 bucket for further analysis.


fauzigo:
A new PR will be submitted

PR #1111, opened 2020-03-02
feat: presigned-url-service for indexd

BinamB:

### New Features
-  Added the base_url for presigned-url-fence-service in indexd_settings for DRS endpoints.

### DIFF
--- a/apis_configs/indexd_settings.py
+++ b/apis_configs/indexd_settings.py
@@ -61,4 +61,6 @@ else:
         )
     )
 
+CONFIG["PRESIGNED_URL_ENDPT"] = "http://presigned-url-fence-service/"
+
 settings = {"config": CONFIG, "auth": AUTH}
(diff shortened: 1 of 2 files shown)


apis_configs/indexd_settings.py
```diff
@@ -61,4 +61,6 @@ def load_json(file_name):
         )
     )
 
+CONFIG["PRESIGNED_URL_ENDPT"] = "http://presigned-url-fence-service/"
+
```
frickjack: that CONFIG variable should default to that value
we don't like to update those settings.py files - we want to retire those
it looks like kube-setup-indexd does update that secret automatically: https://github.com/uc-cdis/cloud-automation/blob/master/gen3/bin/kube-setup-indexd.sh#L22
but I hate that we have .py code in cloud-automation/ that gets inserted into service runtimes
try to setup the service configuration so that it loads all non-secret configuration from configmaps setup out of the manifest folder
and all secret information comes from g3auto/ secrets

PR #1130, opened 2020-03-17
feat(ebs-script): made script to update ebs volumes

emalinowski:

### New Features
Added script to update ebs volumes

### DIFF
--- a/gen3/bin/ebs.sh
+++ b/gen3/bin/ebs.sh
@@ -1,0 +1,218 @@
+#/bin/bash
+
+source "${GEN3_HOME}/gen3/lib/utils.sh"
+gen3_load "gen3/gen3setup"
+
+# lib -----------------------
+
+gen3_ebs_help() {
+  gen3 help ebs
+}
+
+# Lists ebs volumes
+gen3_ebs_list_volumes() {
+  (
+    gen3 aws ec2 describe-volumes
+  )
+}
+
+# Lists ebs snapshots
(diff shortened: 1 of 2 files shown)


gen3/bin/ebs.sh
```diff
@@ -0,0 +1,218 @@
...
+    return 1
+  fi
+  if [[ -z "$volume" ]]; then
+    gen3_log_info "Use: gen3 ebs snapshot volume"
+    return 1
+  fi
+  (
+    set -e
+    check=$(aws ec2 describe-volumes --volume-id $volume| jq -r .Volumes[0])
+    if [[ -z $check ]]; then
```
themarcelor: will the `$check` variable be empty (-z) if the output of the `describe-volumes | jq` json-parsing command doesn't return anything? Perhaps you should try:
```
check=$(aws ec2 describe-volumes --volume-id $volume| jq -r .Volumes[0])
RC=$?
if [[ $RC -ne 0 ]]; then
```

emalinowski: the ec2 describe volumes returns a JSON object. If the Volumes array is empty the jq-r .Volumes[0] will be empty and it will signal there is no volume with that id.

themarcelor: Will it also return an empty JSON block?
If it's an empty string.. wouldn't that break the `-z` check?
e.g.,
```
% echo "" | jq .Volumes[0]
zsh: no matches found: .Volumes[0]
```

PR #1168, opened 2020-04-03
feat: new structure for covid19 etl

m0nhawk:
### Improvements
- Restructure the ETL for COVID-19 Data Commons

### Deployment changes
- All ETL now have the same image and entrypoint, depend on the environmental variable

### DIFF
--- a/files/scripts/covid19-etl-cronjob.sh
+++ b/files/scripts/covid19-etl-cronjob.sh
@@ -1,0 +1,41 @@
+#!/bin/bash
+#
+# Runs daily COVID-19 Illinois Department for Public Health
+# Run as cron job in covid19@adminvm user account
+#
+# USER=USER
+# KUBECONFIG=path/to/kubeconfig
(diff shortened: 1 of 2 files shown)


files/scripts/covid19-etl-cronjob.sh
```diff
@@ -0,0 +1,41 @@
...
+help() {
+  cat - <<EOM
+Use: bash ./covid19-etl-cronjob.sh
+EOM
+}
+
+
+# main -----------------------
+
+if [[ -z "$USER" ]]; then
```
paulineribeyre, line 34: add a similar check for `JOB_NAME`?

m0nhawk: Missed that. Now fixed. 👍

m0nhawk:
> let's remove the 2 idph and jhu specific bash and yaml file?

I want to first deploy the new ETL with this script and only after that remove old ones, just to simplify rollbacks (not mess for git revert etc).

PR #1188, opened 2020-04-14
chore(qa): Do not use selectedTest PR label selection

themarcelor:
Adding an empty string just to comply with the method signature.
This repo does not leverage the PR labelling magic.


themarcelor:
That might not be necessary.

PR #1193, opened 2020-04-16
chore(qa): Trying to restore the pipeline by downgrading Jenkins 

themarcelor:
We are facing lots of intermittent failures so we might need to downgrade to help us narrow down the root cause of all the unstable test runs.


themarcelor:
This is no longer necessary.

PR #1206, opened 2020-04-21
chore(qa): Creating Jenkins2 with Kubernetes plugin capabilities

themarcelor:
This is an experiment based on the Innovation Day project

### DIFF
--- a/Docker/Jenkins2/Dockerfile
+++ b/Docker/Jenkins2/Dockerfile
@@ -1,0 +1,24 @@
+FROM jenkins/jenkins:2.233
+
+USER root
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+# install python and pip and aws cli
+RUN set -xe && apt-get update && apt-get install -y apt-utils dnsutils python python-setuptools python-dev python-pip python3 python3-pip build-essential zip unzip jq less vim gettext-base
+RUN set -xe && python -m pip install awscli --upgrade && python -m pip install pytest --upgrade && python -m pip install PyYAML --upgrade
+RUN set -xe && python3 -m pip install pytest --upgrade && python3 -m pip install PyYAML --upgrade
+RUN set -xe && python -m pip install yq --upgrade && python3 -m pip install yq --upgrade
+
+# update /etc/sudoers
+RUN sed 's/^%sudo/#%sudo/' /etc/sudoers > /etc/sudoers.bak \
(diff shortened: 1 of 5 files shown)


Docker/Jenkins2/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+
+USER root
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+# install python and pip and aws cli
+RUN set -xe && apt-get update && apt-get install -y apt-utils dnsutils python python-setuptools python-dev python-pip python3 python3-pip build-essential zip unzip jq less vim gettext-base
+RUN set -xe && python -m pip install awscli --upgrade && python -m pip install pytest --upgrade && python -m pip install PyYAML --upgrade
+RUN set -xe && python3 -m pip install pytest --upgrade && python3 -m pip install PyYAML --upgrade
+RUN set -xe && python -m pip install yq --upgrade && python3 -m pip install yq --upgrade
```
m0nhawk, line 11: Can this be simplified? Also, why you need multiple Python's versions and packages?

Note, shell can be simplified like this. Put after `FROM`.
```
SHELL ["/bin/bash", "-xe", "pipefail", "-c"]
```

Docker/Jenkins2/jenkins-master-deployment.yaml
```diff
@@ -0,0 +1,355 @@
+{{- if .Capabilities.APIVersions.Has "apps/v1" }}
```
m0nhawk, line 1: This doesn't look like official Jenkins [chart](https://github.com/helm/charts/blob/master/stable/jenkins/templates/jenkins-master-deployment.yaml). Is it some simplified version?

themarcelor: I've forked the original one and made some changes.
This is not one of them, perhaps the chart / tarball I've downloaded with `helm fetch` was from an older version, hence, the mismatch with the one from your link.

Docker/Jenkins2/Dockerfile
themarcelor: This is a copy `ipsi literis` from the existing instruction in the original Jenkins1 Dockerfile.
It should be a `TODO`. But, for now, I guess it's ok to have the first set of artifacts for Jenkins2 officially merged into `cloud-automation`.

PR #1219, opened 2020-04-28
feat(aws_batch): aws batch setup

giangbui:
Jira Ticket: [PXP-xxxx](https://ctds-planx.atlassian.net/browse/PXP-5880)

### New Features
aws batch job for bucket manifest generating tool

### DIFF
--- a/doc/bucket-manifest.md
+++ b/doc/bucket-manifest.md
@@ -14,12 +14,12 @@ The tool can be used to any kind of task that requires batch operations. Users w
 Launch a AWS batch operation job to generate a manfest of the source bucket with provided subnets. The result will be put on the manifest bucket.
 
 ```bash
-  gen3 bucket-manifest create <source bucket> <subnets> <manifest bucket>
+  gen3 bucket-manifest create <source bucket> <manifest bucket>
 ```
 
 Ex.
 ```
-gen3 bucket-manifest create cdistest-giangb-bucket1-databucket-gen3 '["subnet-80d559e4","subnet-2f826072","subnet-e1a390ed"]' giangb-bucket-manifest-test
+gen3 bucket-manifest create cdistest-giangb-bucket1-databucket-gen3 giangb-bucket-manifest-test
 ```
 
 ### status
--- a/doc/bucket-manifest.md
+++ b/doc/bucket-manifest.md
@@ -9,7 +9,7 @@ The tool can be used to any kind of task that requires batch operations. Users w
 
 ## Use
 
-### generate
+### create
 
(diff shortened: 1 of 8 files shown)


tf_files/aws/bucket_manifest/root.tf
```diff
@@ -0,0 +1,163 @@
...
+}
+
+resource "aws_iam_role_policy_attachment" "aws_batch_service_role" {
+  role       = "${aws_iam_role.aws_batch_service_role.name}"
+  policy_arn = "arn:aws:iam::aws:policy/service-role/AWSBatchServiceRole"
+}
+
+resource "aws_security_group" "sample" {
+  name = "aws_batch_compute_environment_security_group"
+
```
fauzigo, line 104: Add a line here with the VPC id. I would just copy line 118 and paste it here.

doc/bucket-manifest.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Use
+
+### create
+
+Launch a AWS batch operation job to generate a manfest of the source bucket with provided subnets. The result will be put on the manifest bucket.
+
+```bash
+  gen3 bucket-manifest create <source bucket> <subnets> <manifest bucket>
+```
```
frickjack, line 18: This is very cool, Giang!
Where are the subnets supposed to come from?  Is it expected we'll run these jobs in the same VPC as the k8s cluster?  If not - how do we spin up a new VPC with subnets and S3 endpoints and no external route?

giangbui: The subnets are from the same VPC which can be refered from `subnet_id`. It is supposed that the job would run in existed VPC and existed subnets. However, I have just made some changes in terraform code to create new VPC, routing table, subnets in 9807ce1e8b5abcb5c945294b993dc2a4db2c88cf. So the command is updated to `gen3 bucket-manifest create <source bucket> <manifest bucket>`

tf_files/aws/batch/root.tf
```diff
@@ -0,0 +1,203 @@
...
+resource "aws_internet_gateway" "gw" {
+  vpc_id = "${aws_vpc.new_vpc.id}"
+
+  tags = {
+    Name = "main"
+  }
+}
+
+resource "aws_subnet" "new_subnet" {
+  vpc_id     = "${aws_vpc.new_vpc.id}"
```
frickjack, line 22: Awesome - thanks for setting up a new VPC.  Is it possible to keep the subnets private, and add VPC endpoints for S3 and ECS?

giangbui: The instances need to download quay image from internet so we need public subnets

PR #1220, opened 2020-04-28
fix(bugs): regex and variables

fauzigo:

### Bug Fixes
- Iam service account creation script wouldn't properly work when checking on role ARN as there was a faulty regex check.
- After the role was created, the trust relationship was malformed  resulting on a non usable service account.


fauzigo:
This issues are already being resolved on another branch `feat/sowerjobs`

PR #1259, opened 2020-05-28
Whitelist idph website for covid-19 etl job

paulineribeyre:
### Improvements
- Added .dph.illinois.gov to Squid whitelist, removed www.dph.illinois.gov


paulineribeyre:
this can't pass jenkins because of Quay

PR #1281, opened 2020-06-16
chore(qa): 20 selenium nodes and 1 chrome instance per node

themarcelor:
We have faced numerous issues today related to:
```(node:5363) UnhandledPromiseRejectionWarning: unknown error: Session [88e1f9fbeef33c2b73233d76cf063007] was terminated due to BROWSER_TIMEOUT
```
Trying to bump up the number of nodes and avoid 2 chrome browsers running on the same pod/container.


themarcelor:
Trying selenium 4 before committing this change.

PR #1286, opened 2020-06-22
chore(qa): Restart Selenium Hub + Nodes every night

themarcelor:
Performing a restart to avoid any "resource leakage".

### DIFF
--- a/files/scripts/qa-restart-selenium.sh
+++ b/files/scripts/qa-restart-selenium.sh
@@ -24,4 +24,5 @@ if [[ -z "$USER" ]]; then
   export USER="$(basename "$HOME")"
 fi
 
-kubectl delete pods $(kubectl get pods | grep selenium | awk '{ print $1 }')
+source "${GEN3_HOME}/gen3/gen3setup.sh"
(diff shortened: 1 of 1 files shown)


files/scripts/qa-restart-selenium.sh
```diff
@@ -0,0 +1,27 @@
...
+  exit 1
+fi
+
+PATH="${PATH}:/usr/local/bin"
+
+if [[ -z "$USER" ]]; then
+  export USER="$(basename "$HOME")"
+fi
+
+kubectl delete pods $(kubectl get pods | grep selenium | awk '{ print $1 }')
```
frickjack, line 27: You need to use `g3kubectl` - look at the other cron jobs ...

frickjack:
```
source "${GEN3_HOME}/gen3/gen3setup.sh"
```

themarcelor: I've pushed a new commit. Thanks for the pointers 🙏

PR #1294, opened 2020-06-26
feat(bucket-replication-job): Added aws batch bucket replication job

emalinowski:
### New Features
Added bucket replication job that uses aws batch

### DIFF
--- a/gen3/bin/bucket-replicate.sh
+++ b/gen3/bin/bucket-replicate.sh
@@ -1,0 +1,229 @@
+#!/bin/bash
+
+source "${GEN3_HOME}/gen3/lib/utils.sh"
+gen3_load "gen3/gen3setup"
+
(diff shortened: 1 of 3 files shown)


gen3/bin/bucket-replicate.sh
```diff
@@ -0,0 +1,229 @@
...
+  gen3 job run bucket-replication-${jobId}
+  gen3_log_info "The job is started. Job ID: ${jobId}"
+
+}
+
+# function to check job status
+#
+# @param job-id
+#
+gen3_replicate_generating_status() {
```
giangbui, line 142:
```suggestion
https://github.com/uc-cdis/cloud-automation/blob/51cd2cfcb2247b156e06bc57bf32ab7b7b2b7789/gen3/bin/gcp-bucket-manifest.sh#L119
```

PR #1324, opened 2020-07-15
Chore/log dna log aggregation

fauzigo:

### New Features
- For CSOC attached commons, logs will now be sent over onto logDNA


tf_files/aws/csoc_common_logging/variables.tf
```diff
@@ -31,3 +31,15 @@ variable "threshold" {
 variable "slack_webhook" {
  default = ""
 }
+
+variable "log_dna_function" {
+  default = "arn:aws:lambda:us-east-1:433568766270:function:logdna_cloudwatch"
+}
```
frickjack, line 37: how does this get setup?
we should have a separate script to publish the logdna lambda, and track that code?

tf_files/aws/modules/common-logging/variables.tf
```diff
@@ -40,3 +40,18 @@ variable "threshold"{
 variable "slack_webhook"{
   default = ""
 }
+
+variable "log_dna_function"{
+  description = "Lambda function ARN for logDNA"
+  default     = ""
```
frickjack, line 46: what happens if I set this variable to empty string `""` ?  will the lambda be smart enough to skip trying to invoke a lambda?  It looks like the lambda function checks for `None` instead of an `arn` or whatever?

tf_files/aws/modules/common-logging/lambda_function.py
```diff
@@ -251,6 +251,32 @@ def nice_it(r_data):
...
+#
+##
+
+def send_to_logDNA(payload):
+
+
+    try:
+        # if there is no threshold, let's not even check
+        if os.environ.get('log_dna_function') is not None:
+            log_dna_function = os.environ.get('log_dna_function')
```
frickjack, line 270: Can we add a check here that the `log_dna_function` looks like an `^arn:`, so we can pass an empty string or whatever through as an environment variable?

tf_files/aws/csoc_common_logging/root.tf
```diff
@@ -13,6 +13,9 @@ module "logging" {
   csoc_account_id  = "${var.csoc_account_id}"
   threshold        = "${var.threshold}"
   slack_webhook    = "${var.slack_webhook}"
```
frickjack, line 15: does this work with `terraform12` ?  maybe do something like this:
* add an optional `gen3Metadata.json` to each `tform/aws/script/` folder 
* modify `gen3 workon` to check if that file exists, and set `GEN3_TERRAFORM_PATH` or whatever to point at terraform 12 if it supports terraform 12
* modify `terraform.sh` to key off the TERRAFORM variable set by `gen3 workon` ...

tf_files/aws/csoc_common_logging/variables.tf
fauzigo: The logDNA function was something created by the security team following logDNA documentation. It is a one time thing only and there is no need to code it. Very unlikely to be modified in the near future unless logDNA releases a new version or something.

tf_files/aws/modules/common-logging/lambda_function.py
fauzigo: validation added in latest commit

tf_files/aws/modules/common-logging/variables.tf
fauzigo: validation added in that latest commit

tf_files/aws/csoc_common_logging/root.tf
fauzigo: I would just dedicate a single JIRA/sprint to make this work. Not as trivial as it sounds.

PR #1346, opened 2020-07-30
feat(encrypted-rds): Added default encryption to rds

emalinowski:
### New Features
Added default encryption to rds databases. Also changed default size to t2.small because encryption is not available for t2.micro instances.

### DIFF
--- a/doc/db.md
+++ b/doc/db.md
@@ -113,9 +113,9 @@ ex:
 gen3 db restore fence fence_20190101.backup
 ```
 
-### gen3 db encrypt account profile
+### gen3 db encrypt profile workspace
 
-Used to encrypt the rds databases. You should use the account/profile from gen3 workon. There is also an extra argument you can add to specify the directory to hold the db snapshots. This should be used for larger db's where the root volume is not large enough to store the snapshots.
+Used to encrypt the rds databases. You should use the profile/workspace from gen3 workon used to setup the vpc/rds. There is also an extra argument you can add to specify the directory to hold the db snapshots. This should be used for larger db's where the root volume is not large enough to store the snapshots. This will take care of updating the databases in the account you run it from but you will need to update the rds endpoints for any other namespaces. Once the new rds instances have been verified to be working you can manually delete them using the aws cli/console.
 
 ```bash
 ex.
--- a/doc/db.md
+++ b/doc/db.md
@@ -112,3 +112,12 @@ Create a new database on the same server as `dbname` owned by the `dbname` owner
 ex:
 gen3 db restore fence fence_20190101.backup
 ```
+
+### gen3 db encrypt account profile
+
+Used to encrypt the rds databases. You should use the account/profile from gen3 workon. There is also an extra argument you can add to specify the directory to hold the db snapshots. This should be used for larger db's where the root volume is not large enough to store the snapshots.
+
+```bash
+ex.
+gen3 db encrypt cdistest emalinowskiv1 </mnt/extravolume>
+```
--- a/gen3/bin/db.sh
+++ b/gen3/bin/db.sh
@@ -562,6 +562,143 @@ gen3_db_restore() {
   fi
 }
 
+# Used to encrypt each db
+gen3_db_encrypt() {
+  # Need the account and profile used in gen3 workon to be able to setup terraform
+  # Can optionally take the dump directory to define where to place the psql dumps. Useful for large databases where we want to store the dumps on an extra ebs volume
+  local account=$1
+  local profile=$2
+  local dumpDir=$3
+  # We want to get the security/parameter groups and subnets from the current rds so we can have it match the old ones
+  local securityGroupId=$(aws rds describe-db-instances --filters '{"Name": "db-instance-id", "Values": ["'$vpc_name'-fencedb"]}' | jq -r .DBInstances[0].VpcSecurityGroups[0].VpcSecurityGroupId)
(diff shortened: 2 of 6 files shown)


doc/db.md
```diff
@@ -112,3 +112,12 @@ Create a new database on the same server as `dbname` owned by the `dbname` owner
 ex:
 gen3 db restore fence fence_20190101.backup
 ```
+
+### gen3 db encrypt account profile
+
+Used to encrypt the rds databases. You should use the account/profile from gen3 workon. There is also an extra argument you can add to specify the directory to hold the db snapshots. This should be used for larger db's where the root volume is not large enough to store the snapshots.
```
frickjack, line 118: Should document this as "profile workspace" to be consistent with the workon documentation - maybe also mention that the workspace should be the one that deploys the VPC and RDS, and that the command will modify the state file, creds in the current namespace, for all 3 db's, and whether the user will need to deallocate the old RDS instances ...

gen3/bin/db.sh
```diff
@@ -562,6 +562,143 @@ gen3_db_restore() {
...
+  local profile=$2
+  local dumpDir=$3
+  # We want to get the security/parameter groups and subnets from the current rds so we can have it match the old ones
+  local securityGroupId=$(aws rds describe-db-instances --filters '{"Name": "db-instance-id", "Values": ["'$vpc_name'-fencedb"]}' | jq -r .DBInstances[0].VpcSecurityGroups[0].VpcSecurityGroupId)
+  local dbSubnet=$(aws rds describe-db-instances --filters '{"Name": "db-instance-id", "Values": ["'$vpc_name'-fencedb"]}' | jq -r .DBInstances[0].DBSubnetGroup.DBSubnetGroupName)
+  local dbParameterGroupName=$(aws rds describe-db-instances --filters '{"Name": "db-instance-id", "Values": ["'$vpc_name'-fencedb"]}' | jq -r .DBInstances[0].DBParameterGroups[0].DBParameterGroupName)
+  gen3_log_info "Taking snapshots. Please check that you have enough disk space and stop if disk space gets filled"
+  gen3_log_info "If the databases are large consider mounting a new volume and adding a third parameter to this command to specify a different snapshot directory"
+
+  if [[ -z $dumpDir ]]; then
```
frickjack, line 579: Maybe just default dumpDir to '.' to avoid the `if` block?

gen3/bin/db.sh
```diff
@@ -562,6 +562,143 @@ gen3_db_restore() {
...
+    gen3 db backup indexd  > indexd-backup.sql
+    gen3 db backup fence  >  fence-backup.sql
+    gen3 db backup gdcapi  > gdcapidb-backup.sql
+    gen3 db backup arborist  > arborist-backup.sql
+    gen3 db backup metadata  > metadata-backup.sql
```
frickjack, line 584: WTS also creates a db if it's deployed

gen3/bin/db.sh
```diff
@@ -562,6 +562,143 @@ gen3_db_restore() {
...
+  local originalFenceDbUrl=$(aws rds describe-db-instances --filters '{"Name": "db-instance-id", "Values": ["'$vpc_name'-fencedb"]}' | jq -r .DBInstances[0].Endpoint.Address)
+  local originalIndexdDbUrl=$(aws rds describe-db-instances --filters '{"Name": "db-instance-id", "Values": ["'$vpc_name'-indexddb"]}' | jq -r .DBInstances[0].Endpoint.Address)
+  local originalGdcApiDbUrl=$(aws rds describe-db-instances --filters '{"Name": "db-instance-id", "Values": ["'$vpc_name'-gdcapidb"]}' | jq -r .DBInstances[0].Endpoint.Address)
+  gen3_log_info "Updating fence db name from $originalFenceDbUrl to $newFenceDbUrl in $(gen3_secrets_folder)"
+  grep -rl $originalFenceDbUrl "$(gen3_secrets_folder)" | xargs sed -i "s/$originalFenceDbUrl/$newFenceDbUrl/g"
+  grep -rl $originalIndexdDbUrl "$(gen3_secrets_folder)" | xargs sed -i "s/$originalIndexdDbUrl/$newIndexdDbUrl/g"
+  grep -rl $originalGdcApiDbUrl "$(gen3_secrets_folder)" | xargs sed -i "s/$originalGdcApiDbUrl/$newGdcApiDbUrl/g"
+
+  gen3_log_info "Disabling gitops sync before updating secrets to ensure services not rolled during db setup"
+  kubectl delete cronjob gitops-sync
```
frickjack, line 630: use `g3kubectl` in scripts - it shouldn't make a difference now, but it used to - might save us some trouble later

gen3/bin/db.sh
```diff
@@ -562,6 +562,143 @@ gen3_db_restore() {
...
+    gen3_log_info "restoring arborist db"
+    gen3_db_psql "fence" -c "CREATE DATABASE \"arborist\";"
+    gen3_db_reset "arborist"
+    gen3 psql arborist  < arborist-backup.sql
+    gen3_log_info "restoring metadata db"
+    gen3_db_psql "sheepdog" -c "CREATE DATABASE \"metadata\";"
+    gen3_db_reset "metadata"
+    gen3 psql metadata  < metadata-backup.sql
+  else
+    gen3_log_info "restoring indexd db"
```
frickjack, line 662: see comment above about defaulting dumpDir to '.' to avoid if/else block

PR #1364, opened 2020-08-12
Feat/slurm

fauzigo:

### New Features
- Slurm cluster module and ansible role


gen3/bin/workon.sh
```diff
@@ -127,5 +153,8 @@ EOM
   fi
 fi
 
-echo "Running: terraform init --backend-config ./backend.tfvars $GEN3_TFSCRIPT_FOLDER/"
-gen3_terraform init --backend-config ./backend.tfvars "$GEN3_TFSCRIPT_FOLDER/"
+tversion=$(checkTerraformModule ${GEN3_TFSCRIPT_FOLDER})
+
+
+echo "Running: terraform${tversion} init --backend-config ./backend.tfvars $GEN3_TFSCRIPT_FOLDER/"
+gen3_terraform${tversion} init --backend-config ./backend.tfvars "$GEN3_TFSCRIPT_FOLDER/"
```
frickjack, line 160: what if we did something like `export GEN3_TFORM_VERSION...`, in `gen3 workon`, then have `gen3_terraform` key on that to run the right terraform version, rather than repeating this check multiple places in the code?

gen3/lib/utils.sh
```diff
@@ -362,3 +362,29 @@ gen3_encode_uri_component() {
     echo "$str"
   )
 }
+
+
+#
+# if the module has a manifest, most likely there is a terraform version
+# value that would help us determine which terraform version to use
+#
+checkTerraformModule() {
```
frickjack, line 371: Why do we have 2 of these?

fauzigo: I think I just renamed the one in `gen3/bin/workon.sh` before removing it because I thought it was the place for the function to be, but at some point, it was not the case so I had to move it to `gen3/lib/utils.sh`

I'll remove the duplication.

PR #1393, opened 2020-09-01
Upgrade helm to helm3

jawadqur:
Jira Ticket: [PXP-6612](https://ctds-planx.atlassian.net/browse/PXP-6612)

### Dependency updates
Upgrade helm to helm3, add a migration path from helm2 -> helm3 for prometheus/grafana if helm2 is already installed.


gen3/bin/kube-setup-workvm.sh
```diff
@@ -165,12 +165,39 @@ EOM
...
+      
+      # delete tiller and other helm2 data/configs
+      helm 2to3 cleanup --skip-confirmation
+    }
+
+    if ! which helm > /dev/null 2>&1; then
+      install_helm
+    else 
+      # Overwrite helm2 with helm3
+      if ! helm version --short | grep v3 > /dev/null 2>&1; then
```
frickjack, line 195: It's a little tricky.  We can have multiple VPC's administered by different user's in the same VM.  So it's possible for something like this to happen:
* vpc1 updates - helm gets updated, and prometheus and grafana get migrated
* vpc2 updated - help has already been updated, so prometheus and grafana don't migrate

You may want to move the migrate out to `kube-setup-tiller` - I assume that doesn't require `sudo` too (most of our accounts don't have sudo), and we want to make sure the migrate happens before we drop tiller probably?

jawadqur: This step is not very idempotent, so I am afraid if we add it to `kube-setup-tiller` we will end up erroring out if the migration has already happened. I need to make it more idempotent somehow, where the migrations only happens if these conditions are met: 

- Client is helm3

- Tiller is still deployed

- Grafana and prometheus deployments exist, but not as a helm release.

PR #1400, opened 2020-09-09
Fix various issues with account-policies module

jawadqur:
### Bug Fixes
Fix various issues with the account-policies tf module


frickjack (approved):
Awesome!

PR #1407, opened 2020-09-17
Fix installation of 2to3 plugin for tiller cleanups

jawadqur:
### Bug Fixes
Fix of a bug when helm 2to3 plugin is not being installed properly and thus `kube-setup-tiller` will fail


fauzigo (approved):
:+1:

PR #1416, opened 2020-09-30
PXP-6813 feat: add env var for iframe app.

mfshao:
This is the accompanying PR of https://github.com/uc-cdis/data-portal/pull/747

Add `iframe_application_urls` in `manifest-portal` which is needed to update portal headers

### New Features
- Add `iframe_application_urls` in `manifest-portal` which is needed to update portal headers


frickjack (approved):
ok - put a comment in slack:

I'll approve this, but you might consider either:
* adding this to gitops.json
* just mounting the whole manifest-portal json key as a file, and extend portal to also load manifest-portal.json or something like that at startup for configuration

PR #1423, opened 2020-10-07
feat(bucket-size-report): Added job to create a bucket size report

emalinowski:
### New Features
Added automation for creating a bucket size report

### DIFF
--- a/doc/bucket-size-report.md
+++ b/doc/bucket-size-report.md
@@ -40,8 +40,8 @@ creds.json
 {
   "credentials": [
     {
-      "AWS_ACCESS_KEY_ID": "", #pragma: allowlist secret
-      "AWS_SECRET_ACCESS_KEY": "" #pragma: allowlist secret
+      "aws_access_key_id": "",
+      "aws_secret_key: ""
     }
   ]
 }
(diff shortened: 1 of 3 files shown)


frickjack (commented):
This is very cool!

* it would be cool to have a gen3 wrapper, so `gen3/bin/bucket-size.sh` just runs `python $GEN3_HOME/gen3/lib/bucket-size/bucket-size.py` - something like that.
* also - add a `gen3/doc/bucket-size.md` help doc - `gen3 help bucket-size` and `gen3 bucket-size help` should both work automagically
* make the SENDER configurable too - can default to you, but allow override in config.json
* it would be good to put some rate limiting on the `process.start()` loop - maybe do batches of 5 or 10 - something like:
```
for batch in batches:
    process_list = for bucketList do process.start() or whateer
    process.join(process_list)
```
* if you're up for it - something I thought would be super cool would be a 
```
gen3 py run
gen3 py install
gen3 py pip
```
and
```
gen3/lib/python/requirements.txt
```
something like that ...
helper commands that run python and pip within a gen3 virtual environment, and list external dependencies in `requirements.txt` (or maybe poetry is better?).  That should make it easier for people to write cloud-automation scripts in python in a consistent way ...

frickjack (approved):
awesome!

PR #1433, opened 2020-10-15
Chore/vpn signed url

frickjack:
* patch vpn new-user script to place deployment package in S3, and generate a signed url to send the user an a one-time secret rather than send the package as an e-mail

### Improvements

* patch vpn new-user script to place deployment package in S3, and generate a signed url to send the user an a one-time secret rather than send the package as an e-mail


fauzigo (approved):
LGTM

PR #1434, opened 2020-10-15
Fix/configmaps del

mfshao:
Update configmap deletion logic to grab existing configmap names from env rather than from manifest file

### Improvements
- Update configmap deletion logic to grab existing configmap names from env rather than from manifest file


jawadqur (approved):
👍

PR #1438, opened 2020-10-23
Pybase: don't wait for wsgi.sock

paulineribeyre:
The block that waits for `wsgi.sock` to appear is before `run uwsgi --ini /etc/uwsgi/uwsgi.ini` so it makes us wait 20 seconds for something that will not happen (since the app hasn't started yet). Remove it doesn't seem to cause issues

### Improvements
- Base python 2 and 3 images: don't wait for wsgi.sock to appear, speed up app start up


frickjack (approved):
Awesome!

PR #1444, opened 2020-10-29
Add appcreds.json to secret on reset

jawadqur:
### Bug Fixes
We have seen jenkins jobs for PR's fail intermittently. After diving into why the jenkins jobs were failing is because the `wts-g3auto` secret didn't contain the appcreds.json and was being mounted as a file instead. 

I am trying to copy whatever is here: 

https://github.com/uc-cdis/cloud-automation/blob/master/gen3/bin/kube-setup-wts.sh#L90


themarcelor (approved):
+1

frickjack (changes requested):
that function is clear_wts_clientID - it's purpose is to clear the client-id out of the WTS secret, since those creds are no longer valid after the fence db has been wiped.   kube-setup-wts will create a new one when gen3 roll all runs.  You can add a comment if you want, but I don't think we want to merge this.

PR #1446, opened 2020-11-09
add new dependency

mattreex:
The gen3 config-env command has a new dependency pygit2. This dependency requires libgit2. 

### Dependency updates
pylib2 added


paulineribeyre (approved):
new `pygit2` dependency requires `python-pygit2` https://github.com/uc-cdis/gen3-release-utils/blob/1b611e6b13371e039a725c24519a7b7ba3354c3f/gen3release-sdk/pyproject.toml#L16

PR #1489, opened 2021-01-05
chore(covid19_job): increase the access token life time

giangbui:
Temporary increase the token lifetime since some project requires much longer to finish for the first run time


jawadqur (approved):

jawadqur:
@paulineribeyre - Should we merge this? I am okay with that if you are.

PR #1493, opened 2021-01-11
task(observability): Incude metrics scraping for services

themarcelor:
Small change to introduce conditional metrics scraping for services that expose the `http://localhost/metrics` endpoint.


themarcelor:
Just testing...

PR #1502, opened 2021-01-15
chore(jenkins): Whitelist jenkins.io to support plugins / security maintenance

themarcelor:
We gotta update some plugins.


fasi-md (approved):
Approved.

PR #1506, opened 2021-01-21
Chore/api key

frickjack:
* `gen3 api api-key username` - generate an api-key for service accounts
* prometheus - bigger disk, lower retention
* cognito - to terraform 12

### Improvements

* `gen3 api api-key username` - generate an api-key for service accounts
* prometheus - bigger disk, lower retention
* cognito - to terraform 12

### DIFF
--- a/kube/services/monitoring/prometheus-values.yaml
+++ b/kube/services/monitoring/prometheus-values.yaml
@@ -706,7 +706,7 @@ server:
 
     ## Prometheus server data Persistent Volume size
     ##
-    size: 110Gi
+    size: 220Gi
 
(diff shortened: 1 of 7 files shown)


kube/services/monitoring/prometheus-values.yaml
```diff
@@ -784,7 +784,7 @@ server:
 
   ## Prometheus data retention period (i.e 360h)
   ##
-  retention: "360h"
+  retention: "240h"
```
themarcelor, line 787: So we will only keep metrics for 10 days, instead of 15 days?
Do we have any growth trends reports on this disk space?

emalinowski (approved):

PR #1507, opened 2021-01-21
fix(tf-squid-module): add cross region support

reisingerf:
When deploying to a different AWS region than `us-east-1` the Terraform `aws_ami` data source cannot find the public AMI (as it is only available in `us-east-1`).
This PR adds a dedicated AWS provider configured to the `us-east-1` region to ensure the AMI can be found even if the deployment (and the base provider) is configured for another region.

It also tries to remove hard coded values by adding variables and using existing parameters.

### DIFF
--- a/tf_files/aws/modules/squid/cloud.tf
+++ b/tf_files/aws/modules/squid/cloud.tf
@@ -53,12 +53,21 @@ resource "aws_iam_instance_profile" "cluster_logging_cloudwatch" {
 ###############################################################
 # AMI
 ###############################################################
+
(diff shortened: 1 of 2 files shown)


reisingerf:
Could I get a quick update on this?
Is this useful, acceptable? Anything from my side I can help with?

jawadqur:
I will merge this into a local branch so I can test all the changes manually before merging to master.

jawadqur:
Please track progress here: https://github.com/uc-cdis/cloud-automation/pull/1790

PR #1513, opened 2021-01-26
(PXP-6471) Cronjob to update visas hourly

BinamB:
Need to get [this](https://github.com/uc-cdis/fence/pull/869) approved first.

Still need to figure out timing such that this cron job doesn't overlap with the usersync. Currently testing in qa environment connected to RAS staging, for around 10 users with visas it takes around 7-8 seconds for this job to run. Would be nice to see how long it would take for more people with bigger visas.

### New Features
- Added cron job to update all user visas hourly.

### DIFF
--- a/kube/services/jobs/fence-visa-update-cronjob.yaml
+++ b/kube/services/jobs/fence-visa-update-cronjob.yaml
@@ -57,8 +57,8 @@ spec:
                   value: /var/www/fence
                 - name: CONCURRENCY
                   GEN3_CONCURRENCY|-value: ""-|
-                - name: WINDOW_SIZE
-                  GEN3_WINDOW_SIZE|-value: ""-|
+                - name: CHUNK_SIZE
+                  GEN3_CHUNK_SIZE|-value: ""-|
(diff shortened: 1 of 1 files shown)


kube/services/jobs/fence-visa-update-cronjob.yaml
```diff
@@ -0,0 +1,139 @@
...
+                    create_command+=" --buffer-size $BUFFER_SIZE"
+                fi
+                if [[ -z "CONCURRENCY" ]]; then
+                    create_command+=" --concurrency $CONCURRENCY"
+                fi
+
+                $create_command
+
+                if [[ $? != 0 ]]; then
+                  echo "Warning: non zero exit code: $?"
```
paulineribeyre, line 135: don't just print a warning; you can do something like [this](https://github.com/uc-cdis/cloud-automation/blob/92e00e33dc51c7fb997914d7d9c9965d25a9b0c0/kube/services/jobs/etl-cronjob.yaml#L92-L93)

paulineribeyre (approved):

PR #1531, opened 2021-02-20
chore(ci): Temporarily disable test_encrypted-rds_workspace

themarcelor:
This is blocking all the PR Checks from `cloud-automation`.

It requires further investigation from the DevOps team.


jawadqur:
These should be fixed in another PR. No need to disable.

PR #1534, opened 2021-02-22
code added to get route tables automatically for VPC peering connection

urvi-occ:
Jira Ticket: [DDI-106](https://occ-data.atlassian.net/browse/DDI-106)

- This script take an optional flag "--get-route-table" to find the route table id of the VPC which is accepting the peering request

### New Features
Added new optional flag "--get-route-table"

### Improvements
when the older script was taking manual entries for route table, this will allow an option to programmatically get route table(s) from the accepter VPC

### DIFF
--- a/CSOC_Documentation.md
+++ b/CSOC_Documentation.md
@@ -63,7 +63,7 @@ A VPC peering connection is used for the communication between the admin VM in t
 
   `gen3 approve_vpcpeering_request <child_vpc_name>`
 
-  The underneath script basically looks for a new VPC peering request, if there is a request in a pending state, it is accepted, tagged and  an appropriate route to the csoc_main_vpc private subnet route table is added.
+  The underneath script basically looks for a new VPC peering request, if there is a request in a pending state, it is accepted, tagged and  an appropriate route to the csoc_main_vpc private subnet route table is added. This scirpt accepts an optional flag `--get-route-table` which programtically gets the route table id(s) of the VPC which is accepting the peering request.
 
 *  Manual Process: Once the VPC Peering connection request is made, the CSOC admin need to do the following (use this process only if `automatic process` explained above give errors):
 	*	Verify the VPC  peering request and accept it. Make sure the VPC peering connection is active. At this time you can give a name to the peering connection for future reference.
--- a/CSOC_Documentation.md
+++ b/CSOC_Documentation.md
@@ -63,7 +63,7 @@ A VPC peering connection is used for the communication between the admin VM in t
 
   `gen3 approve_vpcpeering_request <child_vpc_name>`
 
-  The underneath script basically looks for a new VPC peering request, if there is a request in a pending state, it is accepted, tagged and  an appropriate route to the csoc_main_vpc private subnet route table is added. This scirpt accepts an optional flag `--get-route-table` which programtically gets the route table id(s) of the VPC which is accepting the peering request.
+  The underneath script basically looks for a new VPC peering request, if there is a request in a pending state, it is accepted, tagged and  an appropriate route to the csoc_main_vpc private subnet route table is added. This script accepts an optional flag `--get-route-table` which programmatically gets the route table id(s) of the VPC which is accepting the peering request.
 
 *  Manual Process: Once the VPC Peering connection request is made, the CSOC admin need to do the following (use this process only if `automatic process` explained above give errors):
 	*	Verify the VPC  peering request and accept it. Make sure the VPC peering connection is active. At this time you can give a name to the peering connection for future reference.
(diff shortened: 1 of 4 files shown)


gen3/bin/approve_vpcpeering_request.sh
```diff
@@ -1,23 +1,45 @@
 #!/bin/bash
-if [ $# -ne 1 ]; then
+if [ $# -lt 1 ]; then
     echo "USAGE: $0 name_of_child_vpc"
```
jiaqi216, line 3: Would be helpful to update the Usage and/or add comments on the new flag that's being introduced and how to use it

gen3/bin/approve_vpcpeering_request.sh
```diff
@@ -1,23 +1,45 @@
...
-vpccidrblock=$(aws ec2 describe-vpc-peering-connections --filters Name=status-code,Values=pending-acceptance --query 'VpcPeeringConnections[*].RequesterVpcInfo.CidrBlock' --output text)
 
-aws ec2 accept-vpc-peering-connection --vpc-peering-connection-id $vpcpeerconnid
-aws ec2 create-tags --resources $vpcpeerconnid --tags Key=Name,Value="VPC peering between $child_vpc_name and csoc_main_vpc"
-echo "The vpc peering connection request for id $vpcpeerconnid was accepted"
+    child_vpc_name=$1
+    vpcpeerconnid=$(aws ec2 describe-vpc-peering-connections --filters Name=status-code,Values=pending-acceptance --query 'VpcPeeringConnections[*].VpcPeeringConnectionId' --output text)
+    vpccidrblock=$(aws ec2 describe-vpc-peering-connections --filters Name=status-code,Values=pending-acceptance --query 'VpcPeeringConnections[*].RequesterVpcInfo.CidrBlock' --output text)
+    acceptervpcid=$(aws ec2 describe-vpc-peering-connections --filters Name=status-code,Values=pending-acceptance --query 'VpcPeeringConnections[*].AccepterVpcInfo.VpcId' --output text)
+    ROUTE_TABLES=(rtb-23b6685f rtb-rtb-7ee06301)
```
jiaqi216, line 13: Is this a hard coded route table ID? I guess that's fine since it looks like we've been hardcoding this all along but `rtb-rtb-7ee06301` should probably be `rtb-7ee06301` right?

jiaqi216 (commented):
Left some initial comments but will defer to a devops member for final review.

gen3/bin/approve_vpcpeering_request.sh
urvi-occ: Ya sure. I'll add an readme for it.

gen3/bin/approve_vpcpeering_request.sh
urvi-occ: Yeah the route ID was hardcoded in the previous script so I left it just in case. I am not sure about route table id `rtb-rtb-7ee06301` I guess it is supposed to be `rtb-7ee06301`. It needs to be checked in CTDS CSOC.

CSOC_Documentation.md
```diff
@@ -63,7 +63,7 @@ A VPC peering connection is used for the communication between the admin VM in t
 
   `gen3 approve_vpcpeering_request <child_vpc_name>`
 
-  The underneath script basically looks for a new VPC peering request, if there is a request in a pending state, it is accepted, tagged and  an appropriate route to the csoc_main_vpc private subnet route table is added.
+  The underneath script basically looks for a new VPC peering request, if there is a request in a pending state, it is accepted, tagged and  an appropriate route to the csoc_main_vpc private subnet route table is added. This scirpt accepts an optional flag `--get-route-table` which programtically gets the route table id(s) of the VPC which is accepting the peering request.
```
jiaqi216, line 66: a couple of typos:

`scrirpt` -> `script`
`programtically` ->`programmatically`

gen3/bin/approve_vpcpeering_request.sh
jiaqi216: I think if you preserve what was hardcoded in the script before (i.e. https://github.com/uc-cdis/cloud-automation/pull/1534/files#diff-49fb1a0d6091e1ba2697cbea00f1fc6f8a04004d761459bf0184fe1ed1ef30ecL18)  that would be fine going forward as well. With that said I don't think `rtb-rtb-####` is a valid Route Table ID so it would make sense to correct that

urvi-occ: I considered the way to preserve hardcoded values while coding and tested it so it can work with the hardcoded value when the flag is not passed. About the route table ID, can you help me with getting value of the route table id ?

urvi-occ: I completely agree. I made sure while changing the code in the file to preserve the hardcoded value. If in case the flag is not passed, the script will use the hardcoded values. I also made change for route table ID as before.

jawadqur (approved):
Looks good to me! Thanks for fixing the last typos. I will wait for Jenkins to complete before merging this 👍

CSOC_Documentation.md
urvi-occ: Typos Fixed

PR #1543, opened 2021-03-05
ETL job slack notification

paulineribeyre:
### Improvements
- Slack notification for ETL job or cronjob in case of both success or failure


ocshawn (approved):
looks good to me but i am in no way an expert

jawadqur (approved):

PR #1540, opened 2021-03-04
PXP-7447 Audit service

paulineribeyre:
Jira Ticket: [PXP-7447](https://ctds-planx.atlassian.net/browse/PXP-7447)

audit service repo: https://github.com/uc-cdis/audit-service

### New Features
- Add deployment files for the audit service

PR #1548, opened 2021-03-10
PXP-7447 Roll all: deploy audit-service before fence

paulineribeyre:
https://ctds-planx.atlassian.net/browse/PXP-7447

related to #1540

Fence waits a bit before crashing when audit logs are enabled but it can't reach audit-service, but deploying audit-service last during roll-all makes Fence wait for too long and the Fence pod ends up restarting until audit-service is up.

### Improvements
- Roll all: deploy audit-service before fence


BinamB (approved):
lgtm!

PR #1556, opened 2021-03-23
Add covid19-notebook-etl job

paulineribeyre:
Jira Ticket: [COV-807](https://occ-data.atlassian.net/browse/COV-807)

Goes with https://github.com/uc-cdis/covid19-tools/pull/188

### New Features
- Make copies of the "nb-etl" files; rename them "covid19-notebook-etl". The new "covid19-notebook-etl" job (daily) only runs notebooks. The old files will stay here until we can completely deprecate the "nb-etl" job (weekly), which runs notebooks + the Bayes model. The "nb-etl" job will eventually be replaced by a "covid19-bayes-model" job that only runs the Bayes model.

### DIFF
--- a/kube/services/jobs/covid19-notebook-etl-cronjob.yaml
+++ b/kube/services/jobs/covid19-notebook-etl-cronjob.yaml
@@ -4,8 +4,7 @@ kind: CronJob
 metadata:
   name: covid19-notebook-etl
 spec:
-  # Weekly Thursday 6:00PM Chicago time == Friday 12:00AM UTC
-  schedule: "0 0 * * 5"
+  schedule: "0 10 * * *"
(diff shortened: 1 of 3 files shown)


kube/services/jobs/covid19-notebook-etl-cronjob.yaml
```diff
@@ -0,0 +1,54 @@
+# gen3 job run covid19-notebook-etl S3_BUCKET <S3_BUCKET>
```
jawadqur, line 1: Not a blocker but will this run the job or the cronjob? If it's not the cronjob, we might need to change this comment.

jawadqur (dismissed):

jawadqur (approved):

PR #1573, opened 2021-04-08
fix (AWS EKS module update)

reisingerf:
Remove dependency on hardcoded default AvailabilityZones. Replace
with dynamic lookup from current aws_availability_zones data source.

### Improvements
Removes the dependency of the EKS TF module on the availability_zones hardcoded default values, and the need for custom configuration, while retaining the option to configure the AZs if required.

### DIFF
--- a/tf_files/aws/modules/eks/cloud.tf
+++ b/tf_files/aws/modules/eks/cloud.tf
@@ -4,6 +4,11 @@
 #
 #####
 
+locals{
+  # if AZs are explicitly defined as a variable, use those. Otherwise use all the AZs of the current region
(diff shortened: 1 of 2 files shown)


jawadqur:
Thank you for this @reisingerf - We need to review whether or not this will break existing functionality but this is something we are working towards ourselves. I will let you know if we need to update this PR

jawadqur:
I will merge this into a local branch so I can test all the changes manually before merging to master.

jawadqur:
https://github.com/uc-cdis/cloud-automation/pull/1790

PR #1581, opened 2021-04-20
fix(save-failed-pods-logs): Capture logs before namespace is wiped out

themarcelor:
Capture logs from failed pods when K8sReset fails.

### Bug Fixes
- Capture logs from failed pods when K8sReset fails (before the k8s namespace is teared down).

### DIFF
--- a/gen3/bin/kube-wait4-pods.sh
+++ b/gen3/bin/kube-wait4-pods.sh
@@ -18,25 +18,18 @@ EOM
 }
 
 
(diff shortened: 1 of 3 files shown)


jawadqur (approved):

emalinowski (approved):

haraprasadj (approved):
+1

atharvar28 (approved):
+1

PR #1588, opened 2021-04-27
fix(save-logs): Fetch logs from failed pods when K8sReset fails

themarcelor:
Quick fix to run kubectl logs with both `pod_name` and `container_name`.

### DIFF
--- a/gen3/bin/save-failed-pod-logs.sh
+++ b/gen3/bin/save-failed-pod-logs.sh
@@ -34,7 +34,8 @@ array_of_svc_startup_errors=($(g3kubectl get pods | grep -E "Failed|CrashLoopBac
 for pod in "${array_of_svc_startup_errors[@]}"; do
   pod_name=$(echo $pod | xargs)
   gen3_log_info "storing kubectl logs output into svc_startup_error_${pod_name}.log..."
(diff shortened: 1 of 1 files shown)


gen3/bin/save-failed-pod-logs.sh
```diff
@@ -34,7 +34,8 @@ array_of_svc_startup_errors=($(g3kubectl get pods | grep -E "Failed|CrashLoopBac
 for pod in "${array_of_svc_startup_errors[@]}"; do
   pod_name=$(echo $pod | xargs)
   gen3_log_info "storing kubectl logs output into svc_startup_error_${pod_name}.log..."
-  g3kubectl logs $pod_name > svc_startup_error_${pod_name}.log
+  container_name=$(kc get pod ${pod_name} -o jsonpath='{.spec.containers[*].name}')
```
vpsx, line 37: Just in case: will there always be the `kc` alias set?

vpsx:
thanks for that change! Sorry now I have one more question 😅  Doesn't container_name often end up being something like `first_thing second_thing third_thing` ? I tried some `g3kubectl log`s with multiple args after the -c and it complained at me?

jawadqur (approved):

PR #1590, opened 2021-04-29
Fix/daily metrics capture non prod envs

haraprasadj:
Introduced an optional parameter "ns" to reports-cronjob to enable getting logs mapped to specific namespace. The default value is "default", which does not help capture metrics from vpc's with multiple k8s namespaces like the qa env's.

### DIFF
--- a/files/scripts/reports-cronjob.sh
+++ b/files/scripts/reports-cronjob.sh
@@ -5,10 +5,10 @@
 #
 # PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
 # vpc_name=YOUR-VPC-NAME
-# ns=KUBERNETES_NAMESPACE (optional, defaults to "default")
+# ns_name=K8S_NAMESPACE (optional, defaults to "default")
 # USER=USER
 # KUBECONFIG=path/to/kubeconfig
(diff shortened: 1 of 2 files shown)


files/scripts/reports-cronjob.sh
```diff
@@ -5,6 +5,7 @@
 #
 # PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
 # vpc_name=YOUR-VPC-NAME
+# ns=KUBERNETES_NAMESPACE (optional, defaults to "default")
 # USER=USER
 # KUBECONFIG=path/to/kubeconfig
 # 3   3   *   *   *    (if [ -f $HOME/cloud-automation/files/scripts/reports-cronjob.sh ]; then bash $HOME/cloud-automation/files/scripts/reports-cronjob.sh "vpc=$vpc_name"; else echo "no reports-cronjob.sh"; fi) > $HOME/reports-cronjob.log 2>&1
```
jawadqur, line 11: Can you also update the reports-cronjob to have ns=$ns as arg?

files/scripts/reports-cronjob.sh
```diff
@@ -26,7 +27,7 @@ source "${GEN3_HOME}/gen3/gen3setup.sh"
 
 help() {
   cat - <<EOM
-Use: bash ./reports-cronjob.sh vpc=YOUR_VPC_NAME
+Use: bash ./reports-cronjob.sh vpc=YOUR_VPC_NAME [ns_name=K8S_NAMESPACE]
```
jawadqur, line 30: can you remove the brackets?

jawadqur (approved):

files/scripts/reports-cronjob.sh
haraprasadj: oh, that is just to indicate it is optional :)

jawadqur: 👍

PR #1592, opened 2021-04-30
PXP-8080 fix: disable npm notifier

mfshao:
Jira Ticket: [PXP-8080](https://ctds-planx.atlassian.net/browse/PXP-8080)

Disable npm 7 update notifier because it is causing Portal pods to stuck at starting up after 7 days

Doing this change in cloud-automation here so we don't need to cherry-pick for many old portal images

Will also have a PR for Portal for this fix so it can be applied to future images even if they are not being deployed using cloud-automation 

### Bug Fixes
- Disable npm 7 update notifier because it is causing Portal pods to stuck at starting up after 7 days

### DIFF
--- a/kube/services/portal/portal-deploy.yaml
+++ b/kube/services/portal/portal-deploy.yaml
@@ -86,6 +86,10 @@ spec:
         env:
           - name: HOSTNAME
             value: revproxy-service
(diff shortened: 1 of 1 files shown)


jawadqur (approved):

themarcelor (approved):
+1

PR #1595, opened 2021-05-05
Bump terraform version

jawadqur:
### Dependency updates
Due to a security incident Hashicorp has updated the GPG keys used to sign its releases (including providers and plugins etc) 

That's why we see cloud-automation CI runs fail on Terraform tests. 

We need to bump them to a new version that support the new keys. 

Ref: 
https://discuss.hashicorp.com/t/terraform-updates-for-hcsec-2021-12/23570


themarcelor (approved):
+1

emalinowski (approved):

PR #1597, opened 2021-05-05
flag added to bypass verification of plugins at terraform init

urvi-occ:
Jira Ticket: [OCC-24](https://ctds-planx.atlassian.net/browse/OCC-24)

### Issue
While using `Gen3 workon` command in new linux environment. The following errors terminated the remote s3 connection for terraform. The version of providers used in the script seemed to be the cause for it.

```
Error installing provider "archive": openpgp: signature made by unknown entity.

Error installing provider "aws": openpgp: signature made by unknown entity.

```

### Bug Fixes

to fix this bug, the flag `-verify-plugins=false` is added with terraform init in following places
- gen3/bin/workon.sh
- gen3/bin/refresh.sh

### DIFF
--- a/gen3/bin/refresh.sh
+++ b/gen3/bin/refresh.sh
@@ -66,6 +66,6 @@ fi
 
 echo "Running terraform init ..."
 cd "$GEN3_WORKDIR"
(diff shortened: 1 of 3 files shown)


urvi-occ:
It was fixed this morning. Pull latest cloud automation and run gen3 kube-setup-workvm

PR #1599, opened 2021-05-10
task(fix-it-fridays): Speeding up k8s reset

themarcelor:
Quick experiment


gen3/bin/kube-roll-all.sh
```diff
@@ -79,15 +79,15 @@ fi
 
 if g3k_manifest_lookup .versions.fence 2> /dev/null; then
   # data ecosystem sub-commons may not deploy fence ...
-  gen3 kube-setup-fence
+  gen3 kube-setup-fence &
```
williamhaley, line 82: It it safe to fork some of these setup jobs? I think some apps like fence assume that the audit service ran first to set up the SQS queue, and if things run out of order an app could start with missing resources

gen3/bin/kube-roll-all.sh
```diff
@@ -238,39 +244,41 @@ gen3 kube-wait4-pods || true
...
 if g3k_manifest_lookup .versions.portal 2> /dev/null; then
-  gen3 kube-setup-portal
+  gen3 kube-setup-portal &
 else
   gen3_log_info "not deploying portal - no manifest entry for .versions.portal"
 fi
 
 gen3_log_info "enable network policy"
-gen3 kube-setup-networkpolicy "enable" || true
+gen3 kube-setup-networkpolicy "enable" || true &
```
williamhaley, line 271: I think you may actually want...
```suggestion
{gen3 kube-setup-networkpolicy "enable" || true} &
```
...if I'm understanding this right. Otherwise I think the first command will block.

gen3/bin/kube-roll-all.sh
themarcelor: That's a good point. There must be inter-dependencies between the apps... 
Perhaps the Kubernetes ReadinessProbe should ping the `/status` endpoint and the status check should make sure the other micro-service is up and running -- then the services could be self-aware and we wouldn't need to orchestrate that at the cloud-automation / infra provisioning stage.

gen3/bin/kube-roll-all.sh
themarcelor: cc'iong the original authors:
@m0nhawk @jawadqur

themarcelor: There was no response here.
@williamhaley , do you want to push a commit with your suggestion?

haraprasadj (approved):

PR #1610, opened 2021-05-21
chore(reliability): Bump up nginx rate limit for presigned-url-fence

themarcelor:
Due to the improvements delivered in https://github.com/uc-cdis/fence/pull/774
We can now accommodate more Pre Signed URL requests in our fence pod.


themarcelor:
We shouldn't do this. Closing PR...

PR #1617, opened 2021-05-27
feat(dd-apm): Added initial config for dd apm

emalinowski:

### DIFF
--- a/kube/services/fence/fence-deploy.yaml
+++ b/kube/services/fence/fence-deploy.yaml
@@ -102,6 +102,12 @@ spec:
       - name: fence
         GEN3_FENCE_IMAGE
         env:
+        - name: DD_ENABLED
+          valueFrom:
+            configMapKeyRef:
+              name: manifest-global
+              key: dd_enabled
+              optional: true
         - name: DD_ENV
           valueFrom:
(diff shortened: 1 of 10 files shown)


kube/services/fence/fence-deploy.yaml
```diff
@@ -99,12 +102,40 @@ spec:
...
+        - name: DD_VERSION
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.labels['tags.datadoghq.com/version']
+        - name: DD_LOGS_INJECTION
+          value: "true"
+        - name: DD_PROFILING_ENABLED
+          value: "true"
+        - name: DD_TRACE_SAMPLE_RATE
+          value: "1"
```
jawadqur, line 128: I wish we had proper templating using helm or kustomize so we could omit these whenever DD is not enabled.

jawadqur (approved):

kube/services/fence/fence-deploy.yaml
emalinowski: Yea, true. But it's really just unused env vars at that point right 🤷.

themarcelor (approved):
+1

PR #1618, opened 2021-06-01
installing the latest git version on gen3qa-worker

atharvar28:
updating the git version in Jenkins-worker `Dockerfile`


haraprasadj (commented):
Would be better to merge this with the block above if sudo is not required and block below if sudo is required

Docker/Jenkins-Worker/Dockerfile
```diff
@@ -34,6 +34,11 @@ RUN apt-get update \
      software-properties-common \
      sudo \
   && ln -s /usr/bin/lua5.3 /usr/local/bin/lua
+ 
+# install git 
+RUN apt-get update && \
```
themarcelor, line 39: Looks like it didn't work:
```
6/1/2021, 6:02:14 PM git is already the newest version (1:2.20.1-2+deb10u3). 0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
6/1/2021, 6:02:23 PMRemoving intermediate container ca2b585020da
```
I wonder if this would be more effective:
```
apt-get -t=2.28.0 install git
```

themarcelor (approved):
+1

PR #1622, opened 2021-06-03
deploy thor to gen3 commons kubernetes cluster

jingh8:


kube/services/revproxy/gen3.nginx.conf/thor-service.conf
```diff
@@ -0,0 +1,13 @@
+
+          location /thor/ {
```
themarcelor, line 2: @jingh8 , I think we want to reinforce an authorization header with the environment's access token (or require browser's login to set up the cookie).
I'm trying to dig some documentation on that to help out.

themarcelor (approved):
+1

PR #1629, opened 2021-06-14
chore(ecr): add describe-image command

atharvar28:
Adding ECR `describe-image` to `ecr.sh`

### DIFF
--- a/gen3/bin/ecr.sh
+++ b/gen3/bin/ecr.sh
@@ -200,9 +200,8 @@ gen3_describe_image() {
       gen3_log_err "use: gen3_describe_images repoName tagName"
       return 1
     fi
-    local repo_name="$ecrReg/gen3/$repoName"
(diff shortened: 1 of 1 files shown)


gen3/bin/ecr.sh
```diff
@@ -179,14 +179,25 @@ gen3_ecr_update_policy() {
...
 #
 gen3_ecr_repolist() {
   aws ecr describe-repositories | jq -r '.repositories[] | .repositoryName' | grep '^gen3/'
 }
 
+# Check if the Quay image exists in ECR repository
+gen3_describe_image() {
+    local repoName="$1"
+    shift
+    local imageName="$2"
```
jawadqur, line 193: If this is the docker version tag we should make it clearer by calling it imageTag instead?

atharvar28: I have renamed it to use `tagName` like the whole script uses

jawadqur (approved):

PR #1630, opened 2021-06-14
Fix/gitops sync

emalinowski:
### Bug Fixes
Skipped terraform sqs setup from non-adminvm environments

### DIFF
--- a/gen3/bin/kube-setup-audit-service.sh
+++ b/gen3/bin/kube-setup-audit-service.sh
@@ -89,9 +89,11 @@ if ! g3k_manifest_lookup '.versions["audit-service"]' 2> /dev/null; then
   exit 0
 fi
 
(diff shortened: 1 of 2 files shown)


jawadqur (approved):

paulineribeyre:
Fix gitops-sync issues - details from Ed:
```
The config profile (default) could not be found
INFO: 20:05:48 - Creating SQS '.....'
ERROR: 20:05:49 - PROFILE default not properly configured with default region for aws cli
```
> gitops-sync calls kube-setup commands. The awshelper image doesn’t get setup with an aws creds file, which is what gen3 workon uses to configure terraform with AWS, so it is erroring out trying to run the terraform.

paulineribeyre (approved):

PR #1632, opened 2021-06-15
fix(sa-setup): Added initial sa setup so deployments have access to S…

emalinowski:
### Bug Fixes
Add initial SA setup for that deployments can schedule pods.

### Deployment changes


themarcelor (approved):
+1

PR #1636, opened 2021-06-17
fix: throw errors on missing arguments and fix sed -i logic

atharvar28:
the change will now throw out an error with correct arguments to be passed if the command is missing one of the two arguments when executing the command


themarcelor (approved):
+1

PR #1654, opened 2021-07-09
fix(mariner-creds): Made mariner creds optional so EKS IAM deployment…

emalinowski:
### Bug Fixes
Made mariner creds optional so EKS IAM deployments can leverage the attached SA instead


themarcelor (approved):
+1

PR #1667, opened 2021-07-15
chore(ci): Introduce cli commands to mutate guppy config for CI purposes

themarcelor:
We need to be able to mutate the guppy config in our CI environments for 2 test flows:
 - DYNAMIC data to test the end-to-end flow of the Export to PFB and Export to Workspace features.
 - PREDEFINED data to test Guppy queries based on Queries and Expected Responses (based on Canine DD -- ES index pre-hydrated).

These two utility scripts will be instrumented by gen3-qa tests to help us mutate the k8s configmap to achieve both scenarios.

### DIFF
--- a/gen3/bin/mutate-guppy-config-for-pfb-export-test.sh
+++ b/gen3/bin/mutate-guppy-config-for-pfb-export-test.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+source "${GEN3_HOME}/gen3/lib/utils.sh"
+gen3_load "gen3/gen3setup"
+
+set -xe
+
+# script for mutating the guppy configuration on jenkins env
+# the incoming PR's guppy configuration is mutated to Jenkins environment
+
+# how to run:
+# gen3 mutate-guppy-config-for-pfb-export-test {PR} {repoName}
+
+prNumber=$1
+shift
+repoName=$1
+
+if ! shift; then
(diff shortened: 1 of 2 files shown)


gen3/bin/mutate-guppy-config-for-pfb-export-test.sh
```diff
@@ -0,0 +1,30 @@
...
+shift
+repoName=$1
+
+if ! shift; then
+ gen3_log_err "use: mutate-guppy-config prNumber repoName"
+ exit 1
+fi
+
+kubectl get configmap manifest-guppy -o yaml > original_guppy_config.yaml
+sed -i 's/\(.*\)"index": "\(.*\)_subject",$/\1"index": "'"${prNumber}"'.'"${repoName}"'.\2_subject",/' original_guppy_config.yaml
```
mfshao, line 24: do we still need this index in this case?

mfshao (approved):
👍

gen3/bin/mutate-guppy-config-for-pfb-export-test.sh
themarcelor: Yeah, we need exclusive indices to be utilized by portal-guppy when the PR check mutates into some gitops-qa or cdis-manifest environment's DataDictionary / EtlMapping/ gitops.json, etc.

PR #1677, opened 2021-07-19
feat(sqs-monitoring): Added sqs monitoring

emalinowski:
### New Features
Added sqs monitoring

### DIFF
--- a/gen3/bin/sqs.sh
+++ b/gen3/bin/sqs.sh
@@ -62,6 +62,7 @@ gen3_sqs_create_queue() {
     gen3 cd 1>&2
     cat << EOF > config.tfvars
 sqs_name="$sqsName"
+slack_webhook="$g3k_slack_webhook"
 EOF
     gen3 tfplan 1>&2 || return 1
     gen3 tfapply 1>&2 || return 1
--- a/gen3/bin/sqs.sh
+++ b/gen3/bin/sqs.sh
@@ -62,7 +62,7 @@ gen3_sqs_create_queue() {
     gen3 cd 1>&2
     cat << EOF > config.tfvars
 sqs_name="$sqsName"
-slack_webhook="$g3k_slack_webhook"
+slack_webhook="$(g3k_slack_webhook)"
 EOF
     gen3 tfplan 1>&2 || return 1
     gen3 tfapply 1>&2 || return 1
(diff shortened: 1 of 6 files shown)


gen3/bin/sqs.sh
```diff
@@ -62,6 +62,7 @@ gen3_sqs_create_queue() {
     gen3 cd 1>&2
     cat << EOF > config.tfvars
 sqs_name="$sqsName"
+slack_webhook="$g3k_slack_webhook"
```
paulineribeyre, line 65: how to fix queues that have already been created?

gen3/lib/g3k_manifest.sh
```diff
@@ -122,6 +122,15 @@ g3k_environment() {
...
 g3k_environment > /dev/null 2>&1 || true
 
+#
+# Lookup and cache slack_webhook 
+#
+g3k_slack_webhook() {
+  if [[ -z "$GEN3_SLACK_WEBHOOK" ]]; then
+    GEN3_CACHE_SLACK_WEBHOOK="$(g3kubectl get configmaps global -ojsonpath='{ .data.slack_webhook }')" || return 1
+  fi
+  echo "$GEN3_CACHE_SLACK_WEBHOOK"
```
paulineribeyre, line 132:
```suggestion
  echo "GEN3_CACHE_SLACK_WEBHOOK: $GEN3_CACHE_SLACK_WEBHOOK"
```

gen3/bin/sqs.sh
emalinowski: So I think I would need to rerun the modules manually, but Im not sure if there is a better way to do it.

paulineribeyre: 🤔 i mean if we remember to do it in all the environments that's fine - but i think `kube-setup-fence` has probably been run everywhere by now so that's a lot of environments

emalinowski: Yea, I have to go through and run some other things in every environment anyway 😔.

paulineribeyre: 😨 okay then!

jawadqur (approved):

gen3/lib/g3k_manifest.sh
emalinowski: Thats what ends up getting returned from the function, bash is weird, so we want it to just return the url for the terraform config.

PR #1695, opened 2021-08-02
Feat/hatchery dd apm

mfshao:
Add DD APM support to Hatchery

### New Features
- Add DD APM support to Hatchery

### DIFF
--- a/kube/services/hatchery/hatchery-deploy.yaml
+++ b/kube/services/hatchery/hatchery-deploy.yaml
@@ -20,6 +20,7 @@ spec:
         public: "yes"
         netnolimit: "yes"
         userhelper: "yes"
+        tags.datadoghq.com/service: "hatchery"
         GEN3_DATE_LABEL
     spec:
       affinity:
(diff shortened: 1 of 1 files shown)


kube/services/hatchery/hatchery-deploy.yaml
```diff
@@ -20,6 +20,9 @@ spec:
         public: "yes"
         netnolimit: "yes"
         userhelper: "yes"
+        tags.datadoghq.com/service: "hatchery"
+        tags.datadoghq.com/version: "master"
```
jawadqur, line 24: We shouldn't hardcore version?

mfshao: I think this is just a temp default value and will get substituted when this guy kicks in? https://github.com/uc-cdis/cloud-automation/blob/master/gen3/lib/g3k_manifest.sh#L274

But I'm not 100% sure, my bash is rusty 😅

jawadqur: Ah that makes sense!

jawadqur (approved):

PR #1700, opened 2021-08-05
Adding statsd port to the datadog node agent yaml

atharvar28:
Addition of statsd port to the agent yaml so that we could configure our own customizable metric inside datadog UI


themarcelor (approved):
+1

PR #1711, opened 2021-08-17
feat(save-logs): save logs of all containers in pod

m0nhawk:
### New Features
* `save-failed-pod-logs` now will save both `initContainer` and `container` logs for specific pod


themarcelor (approved):
+1

PR #1723, opened 2021-09-02
DO NOT MERGE (just an idea) = chore(ingress-versioning): Pull nginx revproxy sub-config files from the docker-nginx repo

themarcelor:
This PR depends on https://github.com/uc-cdis/docker-nginx/pull/8/files

Our nginx config is currently in cloud-automation and our microservices code changes sit in their respective repos.
As we don't have monorepo, the code changes that reflect in new behaviors for our RESTful API endpoints are not in sync / not backwards compatible with our Nginx / Reverse Proxy configuration changes that perform the URL Rewrites that will control the ingress of our Gen3 Ecosystem and redirect the inbound requests to the correct services.

With this approach, we will be able to have specific microservices versions coupled with specific docker-nginx (revproxy) versions and avoid awkward conditions to apply different revproxy configs for specific microservices versions.

### DIFF
--- a/gen3/bin/kube-setup-revproxy.sh
+++ b/gen3/bin/kube-setup-revproxy.sh
@@ -68,12 +68,15 @@ if g3k_manifest_lookup .versions.indexd 2> /dev/null; then
   setup_indexd_gateway
 fi
 
+# pick up docker-nginx (revproxy) version from the manifest.json / versions block.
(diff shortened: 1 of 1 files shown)


themarcelor:
Abandoning this PR as the blast radius of such change is too risky.

PR #1725, opened 2021-09-07
fix(indexd-dd): Added netnolimit policy to fix datadog

emalinowski:
### New Features
Added netnolimit for indexd


themarcelor (approved):
+1

PR #1726, opened 2021-09-07
Feat/fence drs

Avantol13:
### Improvements
- route GA4GH DRS Access requests to more performance Fence endpoint over indexd for latest fence image versions

### DIFF
--- a/gen3/lib/utils.sh
+++ b/gen3/lib/utils.sh
@@ -103,6 +103,67 @@ semver_ge() {
   ) 1>&2
 }
 
+# Takes 2 required and 1 optional arguments:
+#   $1 service name
+#   $2 version of service where tests apply >=
+#   $3 version of service where tests apply >=, in monthly release (2020.xx) format
+#
+# ex: isServiceVersionGreaterOrEqual "fence" "3.0.0"
+# or: isServiceVersionGreaterOrEqual "fence" "3.0.0" "2020.01"
+isServiceVersionGreaterOrEqual() {
+  # make sure args provided
+  if [[ -z "$1" || -z "$2" ]]; then
+    return 0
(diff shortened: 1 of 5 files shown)


gen3/lib/utils.sh
```diff
@@ -103,6 +103,67 @@ semver_ge() {
...
 }
 
+# Takes 2 required and 1 optional arguments:
+#   $1 service name
+#   $2 version of service where tests apply >=
+#   $3 version of service where tests apply >=, in monthly release (2020.xx) format
+#
+# ex: isServiceVersionGreaterOrEqual "fence" "3.0.0"
+# or: isServiceVersionGreaterOrEqual "fence" "3.0.0" "2020.01"
+isServiceVersionGreaterOrEqual() {
```
Avantol13, line 113: this logic was mostly copied and slightly modified from a similar feature in gen3-qa integration tests (for only running certain tests above a given version)

BinamB (approved):

Avantol13:
hmm so I think I found the reason for 1 of the failures. The GA4GH DRS integration test is looking for a 403 with a missing token.. but the DRS spec (and general 401 vs 403 usage) implies it should be a 401 (so that's what I changed it to in fence) https://ga4gh.github.io/data-repository-service-schemas/preview/release/drs-1.0.0/docs/#_responses

So that explains the DRS test failure. I still don't understand what's causing the rest of these.. they seem really unrelated - workspace stuff, audit stuff, etc still trying to figure it out

Avantol13:
Better PR to update existing tests: https://github.com/uc-cdis/gen3-qa/pull/699

themarcelor (approved):
+1

PR #1733, opened 2021-09-17
chore(thor): Load Thor database credentials into thor pod

themarcelor:
Mounting .env file containing the database credentials.


haraprasadj (approved):
+1

PR #1738, opened 2021-09-21
feat(datadog): add dd cfgs to the metadata service deployment

Avantol13:
### Improvements
- add datadog env vars to metadata deployment

### DIFF
--- a/kube/services/metadata/metadata-deploy.yaml
+++ b/kube/services/metadata/metadata-deploy.yaml
@@ -25,6 +25,8 @@ spec:
         # for network policy
         netnolimit: "yes"
         tags.datadoghq.com/service: "object-management"
+        GEN3_ENV_LABEL
+        GEN3_METADATA_IMAGE
         GEN3_DATE_LABEL
     spec:
       affinity:


kube/services/metadata/metadata-deploy.yaml
```diff
@@ -24,6 +24,7 @@ spec:
         userhelper: "yes"
         # for network policy
         netnolimit: "yes"
+        tags.datadoghq.com/service: "object-management"
```
emalinowski, line 27: I think you need to add GEN3_ENV_LABEL and GEN3_METADATA_IMAGE, similar to https://github.com/uc-cdis/cloud-automation/blob/master/kube/services/fence/fence-deploy.yaml#L34-L35. Otherwise, it wont get the correct env and version in datadog.

jawadqur: @Avantol13 - Is this still required? What is the reasoning behind calling this "object-management" ?

Avantol13: this service will be generalized in the future to be considered the "Object Management" service - it already houses part of the new obj management API. It is required - I just haven't had time to work on the installation of ddtrace on the MDS side

emalinowski: If we want, we can probably do something similar to the approach I did for the python base image we manage, https://github.com/uc-cdis/cloud-automation/commit/8f475915dd4dba98fc1ec328d93150100af71dce#diff-ce2490d24554732fd8c1ee0ccd0f40d4af65b55a3b526c7d62911b4cac393886R89-R99. Basically we would probably want to have a startup script instead of just running the gunicorn command and that startup script will have a conditional that checks for an env var set within the manifest to see whether to run ddtrace-run or the regular startup command.

PR #1746, opened 2021-10-08
chore(better-k8s-reset-debug): describe pods when k8sReset fails

themarcelor:
describe pods in every failure scenario.


m0nhawk:
👍

PR #1750, opened 2021-10-13
Feat/storage expiration

Avantol13:
This relies on the `feat/passports` branch in Fence (supporting Passports as AuthN/Z for data access) and assumes the release with be Fence 6.0.0 and 2022.02. If that changes, this will be updated. 

### New Features
- New cronjob to remove expired Google Access based on expirations set in Fence

### Deployment changes
- Newer versions of fence + Google enabled = new cronjob to remove temporary Google access from GA4GH Passports


BinamB (approved):
lgtm! :rocket:

PR #1767, opened 2021-10-28
Feat/mariner deploy aws gateway.

UchicagoZchen138:
Jira Ticket: [PXP-8909](https://ctds-planx.atlassian.net/browse/PXP-8909)

### New Features

- mariner persistent volume and persistent volume claim added as part of automatic deployment
- add mariner pod and mariner engine to indexd ingress policy

### DIFF
--- a/kube/services/mariner/mariner-nfs-pv.yaml
+++ b/kube/services/mariner/mariner-nfs-pv.yaml
@@ -1,0 +1,15 @@
+apiVersion: v1
+kind: PersistentVolume
+metadata:
+  name: mariner-nfs-pv
+spec:
+  storageClassName: ""
+  claimRef:
+    name: mariner-nfs-pvc
+  capacity:
+    storage: 10GB
+  accessModes:
+    - ReadWriteMany
+  nfs:
+    server: 172.24.211.63
+    path: "/newstoragegw-poc"
--- a/kube/services/mariner/mariner-nfs-pv.yaml
+++ b/kube/services/mariner/mariner-nfs-pv.yaml
@@ -3,13 +3,11 @@ kind: PersistentVolume
 metadata:
   name: mariner-nfs-pv
 spec:
-  storageClassName: ""
-  claimRef:
-    name: mariner-nfs-pvc
+  persistentVolumeReclaimPolicy: Delete
   capacity:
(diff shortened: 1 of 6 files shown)


kube/services/mariner/mariner-nfs-pv.yaml
```diff
@@ -0,0 +1,13 @@
...
+metadata:
+  name: mariner-nfs-pv
+spec:
+  persistentVolumeReclaimPolicy: Delete
+  capacity:
+    storage: 100Gi
+  accessModes:
+    - ReadWriteMany
+  nfs:
+    server: 172.24.211.63
```
jawadqur, line 12: this should not be hardcoded, but need to get this dynamically, either from manifest or programmatically.

m0nhawk: AWS CLI can return this:
```
aws storagegateway describe-gateway-information --gateway-arn <gateway_arn> | jq -r '.GatewayNetworkInterfaces[].Ipv4Address'
```

The only thing we'll need a gateway ARN somewhere in the manifest, unless we can map S3 bucket to gateway ARN and use that.

m0nhawk: Actually, we can use `aws storagegateway list-file-shares` to list available file shares, use `aws storagegateway describe-nfs-file-shares --file-share-arn-list <file_share_arn>` to get S3 bucket.

Something like this:

```
aws storagegateway describe-gateway-information --gateway-arn \
  $(aws storagegateway describe-nfs-file-shares \
    --file-share-arn-list \
    $(aws storagegateway list-file-shares | jq -r '.FileShareInfoList[].FileShareARN') \
    | jq -r '.NFSFileShareInfoList[] | select(.Path == "<replace_this_by_mount_path>") | .GatewayARN') \
  | jq -r '.GatewayNetworkInterfaces[].Ipv4Address'
```

In this case, `<replace_this_by_mount_path>` equals to `/newstoragegw-poc`.

Also, there may be multiple file share to the same bucket and mount path, so it's probably should be configurable. And there may be multiple IP addresses.

emalinowski (approved):

PR #1774, opened 2021-11-03
fix(installation of cloud-automation into a subfolder) 

uwwint:
This change will allow for cloud-automation to be installed in a subfolder

### DIFF
--- a/gen3/bin/kube-setup-workvm.sh
+++ b/gen3/bin/kube-setup-workvm.sh
@@ -299,7 +299,7 @@ EOF
 
   if ! grep 'GEN3_HOME=' ${WORKSPACE}/.${RC_FILE} > /dev/null; then
     cat - >>${WORKSPACE}/.${RC_FILE} <<EOF
-export GEN3_HOME=${WORKSPACE}/cloud-automation
+export GEN3_HOME=${GEN3_HOME}
 if [ -f "\${GEN3_HOME}/gen3/gen3setup.sh" ]; then
   source "\${GEN3_HOME}/gen3/gen3setup.sh"
 fi


uwwint:
This PR resolves two issues: 

1. put the gen3-cloud automation into a different subfolder, 
2. allow bucket creation outside of us-east-1 such that gen3 kube-setup-ssjdispatcher BUCKET_NAME works outside the us

gen3/bin/kube-setup-workvm.sh
```diff
@@ -299,7 +299,7 @@ EOF
 
   if ! grep 'GEN3_HOME=' ${WORKSPACE}/.${RC_FILE} > /dev/null; then
     cat - >>${WORKSPACE}/.${RC_FILE} <<EOF
-export GEN3_HOME=${WORKSPACE}/cloud-automation
+export GEN3_HOME=${GEN3_HOME}
```
jawadqur, line 302: I think we still need to set this variable, or it will never be set.

gen3/bin/workon.sh
```diff
@@ -113,7 +113,7 @@ if [[ ! -f "$bucketCheckFlag" && "$GEN3_FLAVOR" == "AWS" ]]; then
   }
 EOM
 )
-    gen3_aws_run aws s3api create-bucket --acl private --bucket "$GEN3_S3_BUCKET"
+    gen3_aws_run aws s3api create-bucket --acl private --bucket "$GEN3_S3_BUCKET" --create-bucket-configuration LocationConstraint="$(aws configure get $GEN3_PROFILE.region)"
```
jawadqur, line 116: It would be cleaner to set this here: https://github.com/uc-cdis/cloud-automation/blob/master/gen3/lib/aws.sh#L15 

```
export AWS_DEFAULT_REGION=$(aws configure get "${AWS_PROFILE}.region" || echo us-east-1)
```

But this would require more granular testing to see if it doesn't break existing setups.

gen3/bin/kube-setup-workvm.sh
uwwint: GEN3_HOME is correctly calculated based on location of kube-setup-workvm.sh in the beginning of the script:
https://github.com/uc-cdis/cloud-automation/blob/5715aa4ff7d98184db88ee54f27f92e7305d6f2d/gen3/bin/kube-setup-workvm.sh#L11

Thus there shouldn't be a need to force it into ${WORKSPACE}/cloud-automation

jawadqur:
https://github.com/uc-cdis/cloud-automation/pull/1868 - Created this PR

PR #1789, opened 2021-11-12
fix(esproxy-bug): Fixed bug where multiple esproxy replicas would cau…

emalinowski:
### Bug Fixes
Fixed bug where multiple esproxy replicas would cause check to timeout


m0nhawk (approved):
🎉

PR #1791, opened 2021-11-15
feat(slack-alarms): Added new slack webhook for alarms

emalinowski:
### New Features
Added second webhook for slack alarms channel, to segregate alerts from alarms

### DIFF
--- a/kube/services/jobs/google-verify-bucket-access-group-cronjob.yaml
+++ b/kube/services/jobs/google-verify-bucket-access-group-cronjob.yaml
@@ -123,7 +123,7 @@ spec:
                 valueFrom:
                     configMapKeyRef:
                       name: global
-                      key: slack_webhook
+                      key: slack_alarm_webhook
               - name: gen3Env
                 valueFrom:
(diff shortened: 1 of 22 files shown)


jawadqur (approved):

kube/services/jobs/google-verify-bucket-access-group-cronjob.yaml
```diff
@@ -123,7 +123,7 @@ spec:
                 valueFrom:
                     configMapKeyRef:
                       name: global
-                      key: slack_webhook
+                      key: slack_alarm_webhook
```
jawadqur, line 126: Should we add both for alarm and alert for this one?

kube/services/jobs/usersync-job.yaml
```diff
@@ -102,7 +102,7 @@ spec:
             valueFrom:
                 configMapKeyRef:
                   name: global
-                  key: slack_webhook
+                  key: slack_alert_webhook
```
jawadqur, line 105: Double check if we need both here too

PR #1801, opened 2021-11-30
HP-623 Fix/aws es timeout

mfshao:
Jira Ticket: [HP-623](https://ctds-planx.atlassian.net/browse/HP-623)

Fixing intermittent issue of  `context deadline exceeded (Client.Timeout exceeded while awaiting headers)` from aggMDS sync job

`aws-es-proxy` was changed to use a customized HTTP client from verison 1.0+, and this new client is the cause of throwing these timeout issues

Pass an option `-timeout` to increase its timeout from 15s (default) to 180s, note this option is oly supported from version `1.2+`

### Improvements
- Increase timeout for `aws-es-proxy` client

### Deployment changes
- Envs that is using `AggMDS` should consider to use `aws-es-proxy: 1.2+`


haraprasadj (approved):
+1

PR #1803, opened 2021-12-03
fix(selenium-whitelist): Updated whitelist to include new file locati…

emalinowski:
### Bug Fixes
Updated whitelist to include new file locations needed for selenium


haraprasadj (approved):
+1

PR #1807, opened 2021-12-07
Add 3 data commons to whitelist

jingh8:
Add data commons listed below to whitelist to allow dataguid.org to access
https://github.com/uc-cdis/cdis-manifest/blob/master/dataguids.org/manifest.json#L89-#L104


haraprasadj (approved):
+1

PR #1827, opened 2022-01-27
Feat/landing

mcannalte:
Jira Ticket: https://ctds-planx.atlassian.net/browse/HP-680

### New Features
- service and deployment for frontend framework
- nginx routing behind /ff endpoint


kube/services/frontend-framework/frontend-framework-deploy.yaml
```diff
@@ -0,0 +1,98 @@
...
+          limits:
+            cpu: 2
+            memory: 4096Mi
+        ports:
+        - containerPort: 3000
+        command: 
+          - npm
+          - run
+          - start
+        env:
```
mfshao, line 82: mount `BASE_PATH` env var from here instead of in the dockerfile?

mcannalte: Just seeing this now - good point, I'll switch this

mfshao (approved):
👍

PR #1828, opened 2022-01-27
feat(jobs): add visa update job for fence (cronjob already exists)

Avantol13:
### New Features
- visa update job for fence (cronjob already exists)


kube/services/jobs/fence-visa-update-job.yaml
```diff
@@ -0,0 +1,109 @@
...
+            fi
+            if [[ -z "THREAD_POOL_SIZE" ]]; then
+                create_command+=" --thread-pool-size $THREAD_POOL_SIZE"
+            fi
+            if [[ -z "BUFFER_SIZE" ]]; then
+                create_command+=" --buffer-size $BUFFER_SIZE"
+            fi
+            if [[ -z "CONCURRENCY" ]]; then
+                create_command+=" --concurrency $CONCURRENCY"
+            fi
```
johnfrancismccann, line 97: Should these be `if [[ -n ...` as opposed to `if [[ -z ...`?

When I tested `gen3 job run fence-visa-update`, the logs showed:
```
usage: fence-create update-visas [-h] [--chunk-size CHUNK_SIZE]
                                 [--concurrency CONCURRENCY]
                                 [--thread-pool-size THREAD_POOL_SIZE]
                                 [--buffer-size BUFFER_SIZE]
fence-create update-visas: error: argument --chunk-size: expected one argument
```
I think the command being executed is `fence-create update-visas --chunk-size --thread-pool-size --buffer-size --concurrency`.

Avantol13: Shoot yeah you're right. I copied this from the cronjob so I think both are broken. I'll have to update both and then I'll retest

johnfrancismccann (approved):

PR #1848, opened 2022-02-24
covid19-notebok-etl job: do not overwrite the startup commands

paulineribeyre:
### Improvements
- covid19-notebok-etl job: do not overwrite the startup commands


wangfan860 (approved):
thanks!!

PR #1853, opened 2022-02-25
Add kube-setup-argo

jawadqur:
Add kube-setup-argo
- Setup s3 bucket for argo
- Setup DB for argo
- Deploy argo using helm


gen3/bin/kube-setup-argo.sh
```diff
@@ -0,0 +1,109 @@
...
+
+    if aws s3 ls --page-size 1 "s3://${bucketName}" > /dev/null 2>&1; then
+      gen3_log_info "${bucketName} s3 bucket already exists"
+      # continue on ...
+    elif ! aws s3 mb "s3://${bucketName}"; then
+      gen3_log_err "failed to create bucket ${bucketName}"
+    fi
+
+
+    gen3_log_info "Creating IAM user argo-user"
```
emalinowski, line 56: Can we add $vpc_name or some other identifier to the user, so we can have multiple per account?

PR #1865, opened 2022-03-14
Move pip install ddtrace to Dockerfile

jawadqur:
- Handle enabling Datadog and tracing without runtime installations


emalinowski (dismissed):

Docker/python-nginx/python3.6-alpine3.7/dockerrun.sh
```diff
@@ -91,7 +91,6 @@ if [[ -z $DD_ENABLED ]]; then
   run uwsgi --ini /etc/uwsgi/uwsgi.ini
 ) &
 else
-pip install ddtrace
```
paulineribeyre, line 94: for `Docker/python-nginx/python3.6-alpine3.7` you didn't add ddtrace to the Dockerfile, is that on purpose?

jawadqur: No that is an oversight. I will add now, thanks!

rnerella92 (approved):

emalinowski (approved):

PR #1871, opened 2022-03-17
Add FIPS enabled AMI to TF

jawadqur:
- Add FIPS enabled AMI to TF
- Add SSM policy to workers iam role

### DIFF
--- a/tf_files/aws/modules/eks/cloud.tf
+++ b/tf_files/aws/modules/eks/cloud.tf
@@ -8,6 +8,7 @@ locals{
   # if AZs are explicitly defined as a variable, use those. Otherwise use all the AZs of the current region
   # NOTE: the syntax should improve with Terraform 12
   azs = "${split(",", length(var.availability_zones) != 0 ? join(",", var.availability_zones) : join(",", data.aws_availability_zones.available.names))}"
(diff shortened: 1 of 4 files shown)


emalinowski (approved):

emalinowski (dismissed):

tf_files/aws/modules/eks/cloud.tf
```diff
@@ -529,8 +536,24 @@ resource "aws_launch_configuration" "eks_launch_configuration" {
   }
 }
 
+# Create a new iam service linked role that we can grant access to KMS keys in other accounts
+# Needed if we need to bring up custom AMI's that have been encrypted using a kms key
+resource "aws_iam_service_linked_role" "autoscaling" {
+  aws_service_name = "autoscaling.amazonaws.com"
+  custom_suffix = "ASG"
```
emalinowski, line 543: If there are multiple commons the arn will match for each and lead to tf errors. We should change this to something either random or specific to the common, so that there is not an overlap. Easiest solution is probably ${var.vpc_name}.

jawadqur: You are 100% correct. I will update it!

emalinowski (approved):

PR #1896, opened 2022-04-10
feat(tf-cleanup-script): Added script to clean up tf plugin dirs

emalinowski:
### New Features
Added script to cleanup tf plugin dirs


jawadqur:
Awesome! This will save a lot of disk space. Let's run it in a few envs and think about moving this to a cronjob as well!

PR #1916, opened 2022-04-26
feat pin argo version to prevent upgrade to 0.14.0 which breaks workf…

UchicagoZchen138:
Jira Ticket: NA
### Deployment changes

- pin version of argo since latest upgrade will break workflow submission


gen3/bin/kube-setup-argo.sh
```diff
@@ -208,7 +208,8 @@ if [[ "$ctxNamespace" == "default" || "$ctxNamespace" == "null" ]]; then
 
     helm repo add argo https://argoproj.github.io/argo-helm --force-update 2> >(grep -v 'This is insecure' >&2)
     helm repo update 2> >(grep -v 'This is insecure' >&2)
-    helm upgrade --install argo argo/argo-workflows -n argo -f ${valuesFile}
+    # pin the version of chart to 0.13.1 since the latest version 0.14.0 will cause workflows to not run
+    helm upgrade --install argo argo/argo-workflows -n argo -f ${valuesFile} --version 0.13.1
```
kmhernan, line 212: I think @emalinowski and others would like this to be configurable via our manifest... (please say otherwise)

PR #1923, opened 2022-05-04
feat(kubecost-setup): Added kubecost setup

emalinowski:
### New Features
Kubecost


gen3/bin/awsrole.sh
```diff
@@ -24,6 +24,8 @@ gen3_awsrole_help() {
 function gen3_awsrole_ar_policy() {
   local serviceAccount="$1"
   shift || return 1
+  local namespace="$1"
```
jawadqur, line 27: Can we pull in latest from cloud-auto here? I think we both did the same thing in different PR's

jawadqur (approved):

jawadqur (approved):

gen3/bin/awsrole.sh
emalinowski: Master cloud-auto doesn't seem to have support for using the create function. This should allow it to work, and also let us clean up a bit of the kube-setup-ingress code, so we can just call the gen3 awsrole create function instead. My solution is a bit wordier/explicit though, so Im fine following your format to make changes or just merging mine in and overwriting the changes 🤷‍♂️ .

PR #1927, opened 2022-05-11
Create config-local.js including iframe "keep alive" code

pieterlukasse:
Jira Ticket: [PXP-9641](https://ctds-planx.atlassian.net/browse/PXP-9641)

### New Features
- code to let Atlas send "keep session alive" messages to its parent (when Atlas is running in iframe)

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -1016,13 +1016,13 @@
       {
         "hashed_secret": "6e71f9f2b1e96de5a712f899ed26477ebc260a73",
         "is_verified": false,
-        "line_number": 105,
+        "line_number": 87,
         "type": "Secret Keyword"
       },
       {
         "hashed_secret": "317b889ca9fa8789dc1b85714568b1bdf2c7baf3",
         "is_verified": false,
-        "line_number": 108,
+        "line_number": 90,
         "type": "Secret Keyword"
       }
     ],
--- a/kube/services/ohdsi-atlas/README.md
+++ b/kube/services/ohdsi-atlas/README.md
@@ -100,3 +100,18 @@ stringData:
   logging_level_org_ohdsi: info
   logging_level_org_apache_shiro: info
 ```
+
+## Making changes and redeploying to QA
+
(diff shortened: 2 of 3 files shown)


pieterlukasse:
@alilahrime add this to your code in the main app: 

```javascript
   const EVENT_TYPE = "message"; // from event types https://developer.mozilla.org/en-US/docs/Web/API/EventSource/message_event

    window.addEventListener(EVENT_TYPE, function(event) {
        console.log("got message (event.origin): " + event.origin); // you can use this to check if message is coming from  atlas...
        console.log("got message (event.data): " + event.data);
    });
```

alilahrime (approved):

m0nhawk:
The data-portal is doing the following [here](https://github.com/uc-cdis/data-portal/blob/639fded126936cd767722d076deb00a09f594b17/src/Workspace/index.jsx#L89-L100):
```
  oniframeLoad = (e) => {
    // force workspace iframe acquire focus if it does not have yet
    // to fix the noVNC workspace doesn't respond to keyboard event when came up
    e.target.focus();

    // add event listeners for sessionMonitor timeout
    const iframeContent = e.target.contentDocument;
    if (iframeContent) {
      iframeContent.addEventListener('mousedown', () => sessionMonitor.updateUserActivity(), false);
      iframeContent.addEventListener('keypress', () => sessionMonitor.updateUserActivity(), false);
    }
  }
```

I would expect it can be done similarly, but better to check with other frontend developers on this.

pieterlukasse:
@m0nhawk thanks for sharing this code. Do you think this can be done cross-domain as well? Not sure if our Atlas installation will be on a different domain?

pieterlukasse:
@m0nhawk thanks for the info. For cross-domain scenario I think it has to be implemented like we are proposing (with `postMessage` and an event listener on the other side). See also https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage

m0nhawk (approved):

pieterlukasse:
Are we merging this one? The code seems to be on QA already

m0nhawk (approved):

PR #1933, opened 2022-05-17
Feat/cedar wrapper service

mcannalte:
Jira Ticket: [HP-747](https://ctds-planx.atlassian.net/browse/HP-747)

### New Features
- deployment for cedar wrapper service


kube/services/revproxy/gen3.nginx.conf/cedar-wrapper-service.conf
```diff
@@ -0,0 +1,42 @@
+
+          location /cedar/ {
+
+            if ($csrf_check !~ ^ok-\S.+$) {
+                return 403 "failed csrf check";
+              }
+
+              set $authz_resource "/workspace";
```
mcannalte, line 8: This still needs some revision but could use another set of eyes. I tried copying the authz logic from other endpoints but seems to be getting some different results. 

The goal is to protect authorize this endpoint to anyone with the `/cedar` resource (`/workspace` is a placeholder)

mfshao: what is the different results in here? This looks okay to me as long as we set the resource, method and service, and then pass to `/gen3-authz`

gen3/bin/kube-setup-cedar-wrapper.sh
```diff
@@ -0,0 +1,17 @@
+source "${GEN3_HOME}/gen3/lib/utils.sh"
+gen3_load "gen3/lib/kube-setup-init"
+
+[[ -z "$GEN3_ROLL_ALL" ]] && gen3 kube-setup-secrets
+
+
+cedar_api_key_file="$(gen3_secrets_folder)/cedar_api_key.txt"
+
+if [[ ! -f cedar_api_key_file ]]; then
```
mfshao, line 9:
```suggestion
if [[ ! -f "$cedar_api_key_file" ]]; then
```

gen3/bin/kube-setup-cedar-wrapper.sh
```diff
@@ -0,0 +1,17 @@
...
+    gen3_log_err "No CEDAR api key present in ${cedar_api_key_file}"
+else
+    g3kubectl create secret generic "cedar-service-api-key" --from-file=cedar_api_key.txt=${cedar_api_key_file}
```
mfshao, line 12: need to check existance of this secret and delete it if so, before creating a new one, otherwise this will error out

similiar to https://github.com/uc-cdis/cloud-automation/blob/master/gen3/bin/kube-setup-portal.sh#L38-L41

mfshao (approved):

PR #1939, opened 2022-05-23
dicom-server readinessProbe and livenessProbe

paulineribeyre:
### Improvements
- DICOM server: add readinessProbe and livenessProbe


haraprasadj (approved):
+1

PR #1958, opened 2022-06-14
Fix/make cloud automation configurable

uwwint:
currently cloud-automation is hard coded in all infrastructure. as a result an external provider cannot influence the key material deployed to those machines. This rather large patch exposes the cloud-automation repository as a variable throughout the AWS tf tiles. 

This should be a non-breaking change, as the default still points to uc-cdis


uwwint:
Forgot something here... Ill close and re-open

PR #1960, opened 2022-06-14
feat add argo template

UchicagoZchen138:
Jira Ticket: [PXP-10029](https://ctds-planx.atlassian.net/browse/PXP-10029)

### New Features

- add new env variable for argo template

### DIFF
--- a/kube/services/portal/portal-deploy.yaml
+++ b/kube/services/portal/portal-deploy.yaml
@@ -189,6 +189,12 @@ spec:
             GEN3_DATA_UPLOAD_BUCKET|-value: ""-|
           - name: BASENAME
             value: /portal
+          - name: ARGO_TEMPLATE
+            valueFrom:
+              configMapKeyRef:
+                name: manifest-portal
+                key: ARGO_TEMPLATE
+                optional: true
         volumeMounts:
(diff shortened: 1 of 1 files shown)


kube/services/portal/portal-deploy.yaml
```diff
@@ -189,6 +189,12 @@ spec:
             GEN3_DATA_UPLOAD_BUCKET|-value: ""-|
           - name: BASENAME
             value: /portal
+          - name: ARGO_TEMPLATE
+            valueFrom:
+              configMapKeyRef:
+                name: manifest-portal
+                key: ARGO_TEMPLATE
```
mfshao, line 196: is this variable going to be shared by other services?

if not, why don't just put it into portal config?

rnerella92 (approved):

kube/services/portal/portal-deploy.yaml
UchicagoZchen138: should not be shared by other services, can you point me to an example env var set in portal config @mfshao

UchicagoZchen138: cool got it, thanks for the info.

PR #1963, opened 2022-06-15
patching livenessProbe 

atharvar28:


kube/services/portal/portal-deploy.yaml
```diff
@@ -65,10 +65,16 @@ spec:
             port: 80
           initialDelaySeconds: 60
           periodSeconds: 60
-          timeoutSeconds: 30
+          timeoutSeconds: 60
           # portal sometimes takes a long time to come up ... -
           # has to fetch the dictionary, relay compile, etc
-          failureThreshold: 10
+          failureThreshold: 40
```
paulineribeyre, line 71: could we decrease `periodSeconds` to maybe 30 so we don't have to wait a whole extra minute once it's finally up? and `timeoutSeconds: 60` seems a bit overkill, this is currently allowing 40+ min for the pod to come up. We don't want to wait 40 min when there is a real problem. Maybe we can set `timeoutSeconds` to 30 and `failureThreshold` to 30 to allow a total of 15 min for the pod to come up?

atharvar28: 👍

jawadqur (approved):

paulineribeyre (approved):
lgtm if this works in qa-covid19 😄

PR #1973, opened 2022-06-21
feat(waf-enabled): Enable waf through a flag in the manifest

emalinowski:
### New Features
Enable waf through flag in the manifest


cmlsn (dismissed):
Approved for waf implementation

cmlsn (approved):

PR #1978, opened 2022-06-23
Add Gen3Env env var

nss10:
Jira Ticket: [PXP-100093](https://ctds-planx.atlassian.net/browse/PXP-10093)

### Bug Fixes
* Add missing Gen3Env variable in etl-job.yaml


paulineribeyre (approved):
oh it's already in the cronjob file, but it still doesn't notify 🤔 maybe the cronjobs were setup before the change and should be re-created in prod systems?

PR #2000, opened 2022-07-21
feat: add doc url

mfshao:
In the future we would like to improve this from using `redirect` to `proxy_pass` so we can preserve the URL in browser

### New Features
- specific a `document_url` in `manifest-global` to let `<HOSTNAME>/documentation` redirects to that external doc site


hughestr (dismissed):

haraprasadj (approved):
+1

ajoaugustine:
+1

PR #2027, opened 2022-08-22
Gpe 527

EliseCastle23:
GPE-527

The above request asks that a lifecycle policy be added to the qa argo s3 bucket. We decided this would be a good time to make some improvements to our gen3 s3 create script and terraform module. 

I added a "aws_s3_bucket_lifecycle_configuration" resource in the child "s3-bucket" terraform module. This resource is conditionally created depending on the "count" variable. I set the count variable equal to a new variable called "lifecycle_count". I added this variable to various modules and scripts to ensure it can be utilized and that the default is always "0". 

Additionally, I edited the gen3 kube-setup-argo script to call gen3 s3 create script so the two required Argo buckets no longer need to be created. 

I then adjusted the gen3 s3 create script to allow it to create a bucket and internal bucket for argo. Additionally, if the bucket that is passed into the create function exists- the script will check s3 for a terraform state file. If the state file does not exist, the workspace will be created and the bucket will be imported/ a tf plan and apply will be run. If the state file exists, a tf plan and apply will be run to keep the state up to date.


EliseCastle23:
needs to be moved to terraform 1.0

PR #2029, opened 2022-08-25
Differentiate between a generic and workspace related Unauthorized requests

nss10:
Jira Ticket: [BRH-222](https://ctds-planx.atlassian.net/browse/BRH-222)

### Improvements
* Redirect users to a more appropriate and generic error page for non-workspace related unauthorized requests

### Dependency updates
* This version of cloud-automation expects the portal to have a front end portal changes related to [PR #1067](https://github.com/uc-cdis/data-portal/pull/1067)


kube/services/revproxy/nginx.conf
```diff
@@ -424,6 +424,13 @@ server {
       include /etc/nginx/gen3.conf/gen3ff-as-root/*.conf;
   }
 
+    location @403error {
+      if ($frontend_root_service = "gen3ff") {
+        return 302 https://$host/portal/403error;
+      }
+      return 302 https://$host/403error;
+  }
```
aartivnkt, line 432: @nss10 instead of calling portal here, perhaps display a more generic message?

nss10: I can surely try that, I was hoping if someone can point me to a use case where a non-workspace related 403 error occurs.

PR #2040, opened 2022-09-19
fix(migrate_acl): allow loading and preference of authz mapping from …

Avantol13:
…fence configuration

### New Features
- Pass user.yaml to the indexd authz migration job if available (so job has same authz mapping fence uses)


johnfrancismccann (approved):

jawadqur:
Doing some spring cleaning - Do we still need this?

Avantol13:
> Doing some spring cleaning - Do we still need this?

Yes, this is definitely needed and actually relevant to be fixed for BDC soon (which will need to migrate Index ACL to Authz). Not sure why we didn't push this through years back.

jawadqur (approved):

PR #2052, opened 2022-10-07
chore(revproxy): update nginx config to increase max request body siz…

Avantol13:
…e (manifestservice has no limit and we're running into issues with it being at 1m default)

### Improvements
- update revproxy nginx config to increase max request body size (manifestservice has no limit and we're running into issues with it being at 1m default)


mfshao (approved):
👍

jawadqur (approved):

PR #2054, opened 2022-10-10
GPE-537

EliseCastle23:
adding conditional scale-in protection for workflow ASG


tf_files-1.0/aws/modules/eks/cloud.tf
```diff
@@ -43,6 +43,7 @@ module "jupyter_pool" {
 
 module "workflow_pool" {
   count                         = var.deploy_workflow ? 1 : 0
+  scale_in_protection           = true
```
jawadqur, line 46: Can we use the variable `scale_in_protection` here too and set it to be true by default?

tf_files-1.0/aws/modules/eks/variables.tf
```diff
@@ -198,3 +198,8 @@ variable "enable_spot_instances" {
 variable "enable_on_demand_instances" {
   default = true
 } 
+
+variable "scale_in_protection" {
+  description = "set scale-in protection on ASG"
+  default     = false
```
jawadqur, line 204: Can we set this to be true by default and only use it for workflows?

jawadqur (approved):

PR #2061, opened 2022-10-25
BRH-301: Extend hatchery to launch prismacloud containers

jawadqur:
Jira Ticket: [BRH-301](https://ctds-planx.atlassian.net/browse/BRH-301)

### New Features
- Add ability to launch prismacloud containers using hatchery for security purposes. 

### Deployment changes
- Currently requires to manually update prisma api key in json format under this path in the adminvm 
  - `~/Gen3Secrets/prisma/apikey.json` 
- format has to be the following: 
```
{
  "AccessKeyID": "XXXX",
  "SecretKey": "XXXX"
}
```

This PR extends `kube-setup-hatchery` to create a kubernetes secret with for prisma cloud credentials, and extending the hatchery deployment to look for this secret and use it, if it exists.

### DIFF
--- a/gen3/bin/kube-setup-hatchery.sh
+++ b/gen3/bin/kube-setup-hatchery.sh
@@ -82,6 +82,15 @@ if ! g3kubectl get sa "$saName" -o json | jq -e '.metadata.annotations | ."eks.a
     gen3 awsrole attach-policy "arn:aws:iam::aws:policy/AWSResourceAccessManagerFullAccess" --role-name ${roleName} --force-aws-cli || exit 1
 fi
 
+if [[ -f "$(gen3_secrets_folder)/prisma/apikey.json" ]]; then
+    ACCESSKEYID=$(jq -r .AccessKeyID "$(gen3_secrets_folder)/prisma/apikey.json")
+    SECRETKEY=$(jq -r .SecretKey "$(gen3_secrets_folder)/prisma/apikey.json")
+    if [[ ! -z "$ACCESSKEYID" && ! -z "$SECRETKEY" ]]; then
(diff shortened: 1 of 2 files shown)


gen3/bin/kube-setup-hatchery.sh
```diff
@@ -82,6 +82,15 @@ if ! g3kubectl get sa "$saName" -o json | jq -e '.metadata.annotations | ."eks.a
     gen3 awsrole attach-policy "arn:aws:iam::aws:policy/AWSResourceAccessManagerFullAccess" --role-name ${roleName} --force-aws-cli || exit 1
 fi
 
+if [[ -f "$(gen3_secrets_folder)/prisma/apikey.json" ]]; then
```
mfshao, line 85: curious: why not use `g3auto` to setup this secret?

jawadqur: I just followed what we are already doing for Datadog as that is closely resembling the usecase for this. We could use g3auto as well.

mfshao: oh okay

I think it makes more sense to setup this secrets in `kube-setup-secrets` because it can be used by system-wide services so it is not Hatchery specific, but I'll leave the decision to PEs

mfshao (approved):

gen3/bin/kube-setup-hatchery.sh
jawadqur: Hmm. I think you're right, I'll update this PR.

jawadqur: If we find another usecase for prisma we can move it out of here, so we can get this merged.

PR #2075, opened 2022-11-11
PXP-10268 Add 'fence-delete-expired-clients' job and cronjob

paulineribeyre:
Jira Ticket: [PXP-10268](https://ctds-planx.atlassian.net/browse/PXP-10268)

Goes with https://github.com/uc-cdis/fence/pull/1057

### New Features
- Add 'fence-delete-expired-clients' job and cronjob


kube/services/jobs/fence-delete-expired-clients-cronjob.yaml
```diff
@@ -0,0 +1,114 @@
+---
```
jawadqur, line 1: We don't need separate file for cronjob, this command  can use single definition and use it as  job or cronjob

```
gen3 job cron '* * * *' xxxx
```

jawadqur (approved):

Avantol13 (approved):

PR #2086, opened 2022-11-29
chore(daily-reports): Updated reports to run daily

emalinowski:
### Improvements
Daily 24h report jobs


kmhernan (approved):
This makes sense as the most immediately effective route to go! I tested some of the bash logic and it worked as expected

jawadqur (approved):

PR #2093, opened 2022-12-06
Update filter metadata compoment in HEAL CEDAR ingestion script

mfshao:
Jira Ticket: [HP-989](https://ctds-planx.atlassian.net/browse/HP-989)

### Improvements
- Update filter metadata compoment in HEAL CEDAR ingestion script

### DIFF
--- a/files/scripts/healdata/heal-cedar-data-ingest.py
+++ b/files/scripts/healdata/heal-cedar-data-ingest.py
@@ -1,9 +1,38 @@
 import argparse
-import json
+import sys
 import requests
 import pydash
-import os
 
+FILTER_FIELD_MAPPINGS = {
+    "Study Type.study_stage": "Study Type",
+    "Data.data_type": "Data Type",
+    "Study Type.study_subject_type": "Subject Type",
+    "Human Subject Applicability.gender_applicability": "Gender",
+    "Human Subject Applicability.age_applicability": "Age"
+}
+SPECIAL_VALUE_MAPPINGS = {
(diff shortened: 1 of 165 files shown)


jawadqur (approved):

files/scripts/healdata/heal-cedar-data-ingest.py
```diff
@@ -16,13 +64,13 @@
 
 if not args.directory:
     print("Directory ID is required!")
-    exit(1)
+    sys.exit(0)
```
MichaelLukowski, line 67: These should probably exit with the code 1 as this is an error

MichaelLukowski (commented):
Just one small issue

files/scripts/healdata/heal-cedar-data-ingest.py
mfshao: yeah ideally we should, but since this is a job if we exit with 1 then k8s will retry it, creating a bunch of doomed job pods...

mfshao: I guess maybe I can put it to exit with 1 but update the job retry policy?

MichaelLukowski: That does make sense, that we don't want the job to retry. But I think we would rather see `error` instead of `completed` on the job run. It would make it easier to know something went wrong

MichaelLukowski (approved):
looks good!

PR #1941, opened 2022-05-24
(PXP-9516): Add `metadata-delete-expired-objects` job

nss10:
Jira Ticket: [PXP-9516](https://ctds-planx.atlassian.net/browse/PXP-9516)

### New Features
* Add the `metadata-delete-expired-objects` job, which removes expired objects from `indexd`, `fence` and `Metadata Service`. Set it up by running the `gen3 kube-setup-metadata-delete-expired-objects-cronjob` command

### Deployment changes
* To set up the `metadata-delete-expired-objects` job, run the `gen3 kube-setup-metadata-delete-expired-objects-cronjob` command. Note that the `metadata-delete-expired-objects-job` client must be granted access to `(resource=/mds_gateway, method=access, service=mds_gateway)` and `(resource=/programs, method=delete, service=fence)`

[PXP-9516]: https://ctds-planx.atlassian.net/browse/PXP-9516?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2097, opened 2022-12-07
PXP-9985 Use client credentials instead of API key

paulineribeyre:
Jira Ticket: [PXP-9985](https://ctds-planx.atlassian.net/browse/PXP-9985)

https://github.com/uc-cdis/cloud-automation/pull/1941

Requires https://github.com/uc-cdis/sower-jobs/pull/42

[PXP-9985]: https://ctds-planx.atlassian.net/browse/PXP-9985?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jawadqur:
I have no idea what's going on with this PR, but I see it's not merging to master 👍   Do we need to rebase or something?

paulineribeyre:
@jawadqur what do you mean, what's wrong? the only thing i can see is that jenkins is not running. But it will run before we merge the parent PR anyways

gen3/bin/kube-setup-metadata-delete-expired-objects-cronjob.sh
```diff
@@ -0,0 +1,48 @@
...
+  gen3_log_info "check metadata-delete-expired-objects secret"
+  local secretsFolder="$(gen3_secrets_folder)/g3auto/metadata-delete-expired-objects"
+  if [[ ! -f "$secretsFolder/config.json" ]]; then
+    local hostname=$(gen3 api hostname)
+    gen3_log_info "kube-setup-metadata-delete-expired-objects-job" "creating fence oidc client for $hostname"
+    local secrets=$(g3kubectl exec -c fence $(gen3 pod fence) -- fence-create client-create --client metadata-delete-expired-objects-job --grant-types client_credentials | tail -1)
+    # secrets looks like ('CLIENT_ID', 'CLIENT_SECRET')
+    if [[ ! $secrets =~ (\'(.*)\', \'(.*)\') ]]; then
+        # try delete client
+        g3kubectl exec -c fence $(gen3 pod fence) -- fence-create client-delete --client metadata-delete-expired-objects-job > /dev/null 2>&1
```
nss10, line 19: I didn't completely understand what we are trying to achieve here. I see we were trying to execute `fence-create` to create a client, but then if the secret is not in the format expected, it looks like we just delete the client and recreate again. Something doesn't seem right to me. Are you sure there is no better way than this? Or am I missing something altogether?

nss10 (commented):
I just got a few comments, other than those, the rest looks good to me.

gen3/bin/kube-setup-metadata-delete-expired-objects-cronjob.sh
paulineribeyre: this is a bit of a pattern in cloud-automation scripts - we try to create the client, and if the output is not in the expected format `(client id, client secret)`, we assume it's because the command is returning an error because the client already exists. The fact that we're in this block in the first place means that the config file does not exist, so the creds for this old client have been lost. So we try deleting the old client before creating it. If that doesn't work, then an existing client is not the problem, and we return an error

nss10 (approved):
LGTM

PR #2151, opened 2023-02-13
Feat/argo workflow metrics

AidanHilt:

### Bug Fixes
Added annotations to the Argo workflow controller pod to enable Datadog to read metrics from it.


gen3/bin/kube-setup-argo.sh
```diff
@@ -243,7 +243,7 @@ if [[ "$ctxNamespace" == "default" || "$ctxNamespace" == "null" ]]; then
 
     helm repo add argo https://argoproj.github.io/argo-helm --force-update 2> >(grep -v 'This is insecure' >&2)
     helm repo update 2> >(grep -v 'This is insecure' >&2)
-    helm upgrade --install argo argo/argo-workflows -n argo -f ${valuesFile}
+    helm upgrade --install argo argo/argo-workflows -n argo -f ${valuesFile} --version 0.22.11
```
emalinowski, line 246: It would be nice if we could set that as the default but allow it to be overwritten, similar to [this](https://github.com/uc-cdis/cloud-automation/blob/master/gen3/bin/kube-setup-karpenter.sh#L19).

jawadqur, line 246: So instead of harcoding this version, let's try to read from manifest, with this version as a default. 

See karpenter script

jawadqur (approved):

PR #2156, opened 2023-02-16
Feat/datadog nginx integration

AidanHilt:
### New Features
Updated the revproxy configuration to enable Datadog to scrape metrics using the Nginx integration. This involved adding an annotation to the pod, as well as changing the config of the /nginx_status endpoint to allow connections from outside the pod. This may be a risk, so we'll want to discuss and possibly adjust.


jawadqur:
Not adding the port to the current service, as the current service is a NodePort service.

emalinowski (approved):

PR #2158, opened 2023-02-20
Update cohort-middleware-deploy and argo-wrapper to have optional configs. 

jawadqur:
### Bug Fixes
- Add optional: true for cohort-middleware config
- Add optional: true for argo-wrapper config

This is necessary so pods will start in envs without this config. Services will be misconfigured though, so we need to write automation scripts to set up configurations for these services.


pieterlukasse:
ℹ️  this PR can be reverted once this ticket is solved: https://ctds-planx.atlassian.net/browse/VADC-447

PR #2211, opened 2023-04-10
making cronjob fence-visa-update optional

ajoaugustine:
Jira Ticket: [GPE-891](https://ctds-planx.atlassian.net/browse/GPE-891)

### Improvements
making cronjob fence-visa-update optional

### Deployment changes
Use ENABLE_FENCE_VISA_UPDATE_CRON in the fence manifest (fence-config-public.yaml) to make cronjob fence-visa-update optional. ENABLE_FENCE_VISA_UPDATE_CRON can be set to true or false.
example:
ENABLE_FENCE_VISA_UPDATE_CRON: true
ENABLE_FENCE_VISA_UPDATE_CRON: false
The current job status remains unaffected if ENABLE_FENCE_VISA_UPDATE_CRON is not set.

[GPE-891]: https://ctds-planx.atlassian.net/browse/GPE-891?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gen3/bin/kube-setup-fence.sh
```diff
@@ -84,10 +84,20 @@ if isServiceVersionGreaterOrEqual "fence" "6.0.0" "2022.07"; then
...
+  if [[ "$ENABLE_FENCE_VISA_UPDATE_CRON" == "false" ]]; then
+      echo "Deleting fence-visa-update cronjob"
+      kubectl delete cronjob fence-visa-update
+  elif [[ "$ENABLE_FENCE_VISA_UPDATE_CRON" == "true" ]]; then
+      if ! g3kubectl get cronjob fence-visa-update >/dev/null 2>&1; then
+          echo "fence-visa-update being added as a cronjob b/c fence >= 6.0.0 or 2022.07"
+          gen3 job cron fence-visa-update "30 * * * *"
+      fi
+  else
+      echo "ENABLE_FENCE_VISA_UPDATE_CRON is not set or is null in the configmap manifest-fence. Skipping fence-visa-update cronjob setup."
```
paulineribeyre, line 100: i think we should default to either true or false 🤔 i would rather disable it by default to reduce the noise and so it won't be set up when we don't need it, but since it's current automatically enabled, it would be a breaking change. We would need to coordinate (Commons that need the job would need to add `ENABLE_FENCE_VISA_UPDATE_CRON: true` to the fence config before we merge this PR). @Avantol13 or @BinamB might have an opinion?

BinamB: I would say we should disable it by default too. I think its been running and failing in places its not supposed to be run in.

Avantol13: `GA4GH_PASSPORTS_TO_DRS_ENABLED`

Avantol13: hmm maybe this should be separate though, that only disables the ability to POST passposts to the DRS endpoint..

Avantol13: ultimately I think the logic of whether we actually sync visas is here: https://github.com/uc-cdis/fence/blame/master/fence/blueprints/login/ras.py#L47-L59

Avantol13: so if we end up adding a new config, we need to make sure that we error with weird combinations like: the above forcing global visa sync but the new config is false, so the cron would never run. That would be really really bad

paulineribeyre: @Avantol13 Here is a Fence PR to add the `ENABLE_VISA_UPDATE_CRON` setting + validation around visa parsing: https://github.com/uc-cdis/fence/pull/1087

paulineribeyre (commented):
lgtm

but we can't merge until the Fence change is merged and the systems that need visa parsing add the `ENABLE_VISA_UPDATE_CRON` config AND deploy the new version of fence

paulineribeyre (approved):
@k-burt-uch confirmed that BDC and DCF do not need config updates

PR #2218, opened 2023-04-17
Feat/vhdc prod workflows scaleup

AidanHilt:
This pull request is to add changes we made in vhdcperf, so they can be deployed in vhdcprod. This covers only the Argo Helm chart and cluster-autoscaler changes, as the Opencost changes have not been finalized. (We want to come to a final decision on sizing for the pods and nodes, and run a smoke test to validate that the setup can survive a very high number of nodes in the clsuter).

### DIFF
--- a/gen3/bin/kube-setup-autoscaler-for-large-workflows.sh
+++ b/gen3/bin/kube-setup-autoscaler-for-large-workflows.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+# Set the resources block for the deployment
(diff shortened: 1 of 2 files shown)


gen3/bin/kube-setup-autoscaler-for-large-workflows.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
```
jawadqur, line 1: Will this script be overriden by gitops job in any way? 🤔

jawadqur (approved):

PR #2221, opened 2023-04-21
Disabling support for TLS Versions 1.0 and 1.1

EliseCastle23:
We would like to only support TLS versions 1.2 or higher per GPE-900.


EliseCastle23:
making a cleaner PR

PR #2230, opened 2023-04-28
Feat/datadog aurora monitoring

AidanHilt:


gen3/bin/kube-setup-aurora-monitoring.sh
```diff
@@ -0,0 +1,169 @@
...
+
+  kubectl patch applications.argoproj.io datadog-application --type merge -n argocd --patch-file /tmp/confd.yaml
+
+else
+  gen3_log_info "We didn't detect an ArgoCD application named 'datadog-application,' so we're going to reinstall the DD Helm chart"
+  
+  (cat kube/services/datadog/values.yaml | yq --arg endpoints "$postgresString" --yaml-output '.clusterAgent.confd."postgres.yaml" = $endpoints | .clusterChecksRunner.enabled = true') > $(gen3_secrets_folder)/datadog/datadog_values.yaml
+  helm repo add datadog https://helm.datadoghq.com --force-update 2> >(grep -v 'This is insecure' >&2)
+  helm repo update 2> >(grep -v 'This is insecure' >&2)
+  helm upgrade --install datadog -f "$(gen3_secrets_folder)/datadog/datadog_values.yaml" datadog/datadog -n datadog --version 3.6.4 2> >(grep -v 'This is insecure' >&2)
```
emalinowski, line 168: Should we just call kube-setup-datadog, so that we only have one place where we manage pinned DD versions and changes?

AidanHilt: We totally could, but that would require changing how kube-setup-datadog works, since this is referring the values file in a different place. I have a story out to change how we handle secrets/sensitive info for Helm charts as a whole, I think we should probably discuss that as a whole team to decide how far we want to tear apart code we're interested in deprecating.

jawadqur (approved):

PR #2231, opened 2023-05-01
Feat/kubecost spot instance feed

emalinowski:
### New Features
Added spot instance feed

### Improvements
Updated kubecost to use more simplified standard deployment


bwinsl0w:
please add both llilly@uchicago.edu and bwinslow@uchicago.edu to kubecost

jawadqur (approved):

PR #2232, opened 2023-05-01
HP-1117 Feat/cedar ingest update

MichaelLukowski:
Jira Ticket: [HP-1117](https://ctds-planx.atlassian.net/browse/HP-1117)
### New Features
Modify cedar ingestion script to paginate cedar wrapper service. We had previous issues with the requests timing out.

[HP-1117]: https://ctds-planx.atlassian.net/browse/HP-1117?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


files/scripts/healdata/heal-cedar-data-ingest.py
```diff
@@ -79,53 +79,84 @@ def update_filter_metadata(metadata_to_update):
...
-            mds_cedar_register_data_body["_guid_type"] = "discovery_metadata"
-
-            print("Metadata is now being registered.")
-            mds_put = requests.put(f"http://revproxy-service/mds/metadata/{cedar_record_id}",
-                headers=token_header,
-                json = mds_cedar_register_data_body
+
+            cedar_appl_id = str(cedar_record["appl_id"])
+            # have to query the mds to find the cedar record id
+            query = requests.get(f"http://revproxy-service/mds/metadata?gen3_discovery.appl_id={cedar_appl_id}")
```
mfshao, line 110: can we use something like `query = requests.get(f"http://revproxy-service/mds/metadata?data=True&gen3_discovery.appl_id={cedar_appl_id}")` to get the entrie metadata block back via one single query rather than query MDS twice?

MichaelLukowski: Added this to the script

files/scripts/healdata/heal-cedar-data-ingest.py
```diff
@@ -79,53 +79,82 @@ def update_filter_metadata(metadata_to_update):
...
+            if mds.status_code == 200:
+                mds_res = mds.json()
+
+                # the query result key is the record of the metadata. If it doesn't return anything then our query failed.
+                if len(list(mds_res.keys())) == 0:
+                    print("Query returned nothing for ", cedar_appl_id, "appl id")
+                    continue
+
+                # get the key for our mds record
+                cedar_record_id = list(mds_res.keys())[0]
```
mfshao, line 122: better also check `if len(list(mds_res.keys())) > 1` before doing this just for sanity check. It shouldn't happen, but if it does, then it probably means we have some issue with our metadata and we should investigate

PR #2238, opened 2023-05-03
GPE-946

EliseCastle23:
### Improvements
Pod disruption budgets will help increase our availability by making sure we always have at least 1 replica available at all times. We are also expanding this to ensure that every services has at least 2 replicas, so if a pod gets evicted the service will remain running.


EliseCastle23:
closing this PR since it seems to be corrupted and I can no longer push to it.

PR #2255, opened 2023-05-22
feat(manifestservice-sa): Updated manifestservice to use EKS IAM SA

emalinowski:
### New Features
Updated manifestservice to use EKS IAM SA


emalinowski:
GPE-950

jawadqur (approved):

PR #2285, opened 2023-07-13
feat: increase proxy buffer size for headers

m0nhawk:
Jira Ticket: N/A

### New Features
* bigger headers size for default python image with nginx


Avantol13:
This looks good but we should get it passing in CI

EliseCastle23 (approved):

PR #2295, opened 2023-07-24
fix(sqs-helper): Updated sqs helper script to create more than one workspace GPE-998

emalinowski:
### Bug Fixes
Updated sqs helper script to create more than one terraform workspace


emalinowski:
GPE-998

jawadqur (dismissed):

PR #2375, opened 2023-10-05
adding sleep to reset.sh

EliseCastle23:


jawadqur:
do we still want this? and may I ask why we're adding sleep? Maybe we can close out if we don't really need a sleep

PR #2377, opened 2023-10-05
GPE-1122 adding logic to exit a gitops sync if the new json contains invalid json

EliseCastle23:
Tested manually

### Improvements
gitops sync will fail if there is a json syntax error anywhere in the manifest.json

### DIFF
--- a/gen3/bin/gitops.sh
+++ b/gen3/bin/gitops.sh
@@ -291,9 +291,15 @@ gen3_gitops_sync() {
   if g3kubectl get configmap manifest-versions; then
     oldJson=$(g3kubectl get configmap manifest-versions -o=json | jq ".data")
   fi
-  newJson=$(g3k_config_lookup ".versions")
   echo "old JSON is: $oldJson"
-  echo "new JSON is: $newJson"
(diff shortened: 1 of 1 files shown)


gen3/bin/gitops.sh
```diff
@@ -291,9 +291,15 @@ gen3_gitops_sync() {
   if g3kubectl get configmap manifest-versions; then
     oldJson=$(g3kubectl get configmap manifest-versions -o=json | jq ".data")
   fi
-  newJson=$(g3k_config_lookup ".versions")
   echo "old JSON is: $oldJson"
-  echo "new JSON is: $newJson"
+  newJson=$(g3k_config_lookup ".versions")
```
paulineribeyre, line 295: I think this update will make sure it fails if any part of the manifest is invalid json, not just the "versions" section, if that's the case could you update the PR description

paulineribeyre (approved):

gen3/bin/gitops.sh
EliseCastle23: yes! Thanks so much Pauline :D Just updated the description

emalinowski (approved):

PR #2381, opened 2023-10-11
fix: skip study if it is neither registered nor unregistered

mfshao:
### Bug Fixes
- HEAL CEDAR ingestion script: ignore study metadata record that is neither unregistered nor registered


george42-ctds (approved):
Nice! Looks good.

PR #2406, opened 2023-11-06
feat: add app-config.js for ohif-viewer

m0nhawk:
Jira Ticket: N/A

### Improvements
* add app-config.js needed for OHIF Viewer with backward compatible configuration to support old and new Orthanc server based on if new Orthanc is deployed in the manifest

### DIFF
--- a/gen3/bin/kube-setup-dicom.sh
+++ b/gen3/bin/kube-setup-dicom.sh
@@ -83,19 +83,17 @@ EOM
 EOM
   fi
 
-  if [[ ! -f "$secretsFolder/app-config.js" ]]; then
-    if g3k_manifest_lookup .versions["dicom-server"] > /dev/null 2>&1; then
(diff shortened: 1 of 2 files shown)


gen3/bin/kube-setup-dicom.sh
```diff
@@ -82,6 +82,21 @@ EOM
 }
 EOM
   fi
+
+  if [[ ! -f "$secretsFolder/app-config.js" ]]; then
```
paulineribeyre, line 86: The `app-config.js` in this repo will only be used for the 1st deployment. Instead should we ensure that every time we run `kube-setup-dicom`, we pull the config from cloud-automation? This way we can maintain the config here instead of editing directly in the vm?

m0nhawk: I can remove the check, so every time it will create from the latest cloud-automation.

paulineribeyre (approved):

PR #2452, opened 2024-01-31
Update migrate to VPC CNI script 

EliseCastle23:
adding coredns update and changing around the order in which vpc cni is updated and nodes are rolled


emalinowski (approved):

jawadqur:
Not going to use VPC cni for network policies, Closing.

PR #2490, opened 2024-02-29
Update heal-cedar-data-ingest.py

mfshao:


mfshao:
we only need awshelper image

PR #2499, opened 2024-03-12
Split build-push workflows

paulineribeyre:
Split the workflow in 3 so that each image only gets built when changes to the corresponding files are detected.
This way we will not run unnecessary workflows and we will not create unused ECR tags, since there is a limit on how many images we can have in ECR.


AidanHilt (approved):

.github/workflows/build_awshelper.yaml
```diff
@@ -0,0 +1,21 @@
+name: Build awshelper image
+
+on:
+  push:
+    paths:
+    - .github/workflows/build_awshelper.yaml
+    - Docker/awshelper/**
```
mfshao, line 7: sorry, late to the game, but I think only triggers awshelper build on these paths are not enough, since there are many jobs that depends on the awshelper image, and the changes to these job can actually happen in locations other than these two. For example https://github.com/uc-cdis/cloud-automation/blob/master/kube/services/jobs/cedar-ingestion-job.yaml#L57

mfshao (commented):
should we uncheck these 3 actions as required from the branch protection rule?

.github/workflows/build_awshelper.yaml
paulineribeyre: I don't think i understand - if you modify `cedar-ingestion-job.yaml` but not the awshelper image, the updated `cedar-ingestion-job` can use awshelper:master without a problem, right? You don't need to use awshelper:<your branch> unless you modified the image?

mfshao: that job uses a file from [here](https://github.com/uc-cdis/cloud-automation/blob/master/files/scripts/healdata/heal-cedar-data-ingest.py) so unless a new awshelper image is being build when changes to that file is merged, then the changes will not always be reflected in the master awshelper image right?

paulineribeyre: oh i see, i'll send another PR

paulineribeyre: thank you!!

PR #2506, opened 2024-03-18
Update ingress.yaml

emalinowski:
### Bug Fixes
Update tls cipher suites


cmlsn (approved):
Approved adding more cipher suites back.

PR #2515, opened 2024-03-28
MIDRC-672 Fix ECR access job role name conflict

paulineribeyre:
Link to JIRA ticket if there is one: https://ctds-planx.atlassian.net/browse/MIDRC-672

### Bug Fixes
- ECR access job: fix conflict during role creating when the role name already exists for a different namespace in the same AWS account

### DIFF
--- a/gen3/bin/iam-serviceaccount.sh
+++ b/gen3/bin/iam-serviceaccount.sh
@@ -115,7 +115,7 @@ EOF
 # @return the resulting json from awscli
 ##
 function create_role(){
-  local role_name="${vpc_name}-${SERVICE_ACCOUNT_NAME}-role"
+  local role_name="${1}"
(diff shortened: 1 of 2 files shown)


gen3/bin/iam-serviceaccount.sh
```diff
@@ -115,16 +115,16 @@ EOF
 # @return the resulting json from awscli
 ##
 function create_role(){
-  local role_name="${vpc_name}-${SERVICE_ACCOUNT_NAME}-role"
+  local role_name="${1}"
```
paulineribeyre, line 118: the role name was already an argument, it just wasn't used: https://github.com/uc-cdis/cloud-automation/blob/251283c6951124730821b6650508de5945c1e259/gen3/bin/iam-serviceaccount.sh#L205

ajoaugustine (approved):

PR #2526, opened 2024-04-19
fix: skip setting up cedar-ingest client cred in ci

mfshao:
### Improvements
- Skip setting cedar ingestion client creds in CI


haraprasadj (dismissed):

haraprasadj (approved):
+1

PR #2552, opened 2024-05-17
Strict cors configuration

jawadqur:
- https://ctds-planx.atlassian.net/browse/GPE-1305 

### Breaking Changes
- Stricter cors configuration to only allow cors from whatever hostname being used to call the revproxy.


jawadqur:
Not needed for now. Closing

PR #2556, opened 2024-05-30
HP-1520 Fix/ct gov

mfshao:
Link to JIRA ticket if there is one: https://ctds-planx.atlassian.net/browse/HP-1520

To support clinicaltrials.gov API V2 for HEAL CEDAR ingestion job

### Improvements
- HEAL CEDAR ingestion: don't update clinicaltrials.gov from CEDAR values

### DIFF
--- a/files/scripts/healdata/heal-cedar-data-ingest.py
+++ b/files/scripts/healdata/heal-cedar-data-ingest.py
@@ -99,7 +99,7 @@ def update_filter_metadata(metadata_to_update):
     filter_metadata = pydash.uniq(filter_metadata)
     metadata_to_update["advSearchFilters"] = filter_metadata
     # Retain these from existing tags
-    save_tags = ["Data Repository", "Common Data Elements"]
+    save_tags = ["Data Repository", "Common Data Elements", "RequiredIDP"]
     tags = [tag for tag in metadata_to_update["tags"] if tag["category"] in save_tags]
     # Add any new tags from advSearchFilters
     for f in metadata_to_update["advSearchFilters"]:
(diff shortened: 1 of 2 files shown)


files/scripts/healdata/heal-cedar-data-ingest.py
```diff
@@ -114,7 +114,7 @@ def update_filter_metadata(metadata_to_update):
 
 def get_client_token(client_id: str, client_secret: str):
     try:
-        token_url = f"http://revproxy-service/user/oauth2/token"
+        token_url = "http://revproxy-service/user/oauth2/token"
```
MichaelLukowski, line 117: Can we change this to `https`? Snyk is throwing this url as an error. I know that this is hitting reverse proxy but I think that https will still probably work.

mfshao: Yeah I wasn't sure if https would work, I was going to tell security this is a false-positive. But I can definitely test the https change

mfshao: well, actually, it doesn't look like https gonna work for revproxy 😕 https://github.com/uc-cdis/cloud-automation/blob/master/kube/services/revproxy/revproxy-service.yaml#L8-L12

MichaelLukowski: Yeah, had a small chat with Jawad about this and it won't work as handling certs for all the services would be a huge issue we would have to maintain.

MichaelLukowski (approved):
Everything looks good! Thanks for moving everything to the correct service rather than the revproxy.

PR #2570, opened 2024-06-18
HP-1521 Feat/refresh ctgov metadata

mfshao:
Link to JIRA ticket if there is one: https://ctds-planx.atlassian.net/browse/HP-1521

### Improvements
- HEAL CEDAR ingestion: this script will now query clinicaltrials.gov API and pull latest metadata from there during each run


files/scripts/healdata/heal-cedar-data-ingest.py
```diff
@@ -50,6 +51,50 @@
     "BioSystics-AP": "https://biosystics-ap.com/assays/assaystudy/<STUDY_ID>/",
 }
 
+CLINICAL_TRIALS_GOV_FIELDS = [
```
mfshao, line 54: we should change this from being hard-coded here to mount it in via k8s cm

george42-ctds (approved):
Looks good.

MichaelLukowski (approved):
looks good to me

PR #2575, opened 2024-06-20
feat: remove Atlas/WebAPI configuration, not needed, it's exposed via ALB

m0nhawk:
Link to JIRA ticket if there is one: N/A

### Improvements
* removing unneeded nginx configuration


pieterlukasse (approved):
nice. Where is the real setting to be found then?

m0nhawk:
@pieterlukasse Atlas service is exposed directly via ALB, so this: https://github.com/uc-cdis/Atlas/blob/2.15.0-DEV/docker/nginx-default.conf

PR #2588, opened 2024-07-10
update squid whitelist

ajoaugustine:
update squid whitelist


MichaelLukowski (approved):
thanks for making the change so fast!

PR #2589, opened 2024-07-11
Feat: set webapi token exp. to 15 min and refresh when 4 min left

pieterlukasse:
Link to JIRA ticket if there is one: https://ctds-planx.atlassian.net/browse/VADC-1241

### New Features
- feat: set webapi token expiration to 15 min and refresh when 4 min left to simulate a session timeout (inactivity timeout) of 15 min


m0nhawk (approved):
👍

PR #2593, opened 2024-07-16
Adding an alert to all workflows if they get timed out

AidanHilt:


m0nhawk (approved):
Some small notes:
* It would be nice to use some ready-to-use Docker images with `jq` and `curl`, instead of installing this in runtime.
* Document this, otherwise it'll fail if the secret is not present.

Otherwise good change. 👍

m0nhawk (approved):
:+1:

PR #2604, opened 2024-07-25
Feat/backup script

ajoaugustine:
install mount-s3

### DIFF
--- a/gen3/bin/dbbackup.sh
+++ b/gen3/bin/dbbackup.sh
@@ -61,6 +61,7 @@ create_or_get_kms_key() {
     gen3_log_info "KMS key with alias $kms_key_alias already exists"
   fi
   kms_key_arn=$(aws kms describe-key --key-id $kms_key_id --query "KeyMetadata.Arn" --output text)
+  gen3_log_info "KMS Key ARN: $kms_key_arn"
 }
 
 # Create an S3 access policy if it doesn't exist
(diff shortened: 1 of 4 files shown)


emalinowski (approved):

gen3/bin/dbbackup.sh
```diff
@@ -222,28 +224,289 @@ copy_to_aurora() {
...
+EOF
+  fi
+}
+
+# Check prerequisites for encrypted backup and cronjob
+check_prerequisites() {
+    create_or_get_kms_key
+    create_s3_bucket $bucket_name $kms_key_arn
+    create_s3_bucket $bucket_name_encrypted $kms_key_arn
+    setup_csi_driver
```
jawadqur, line 482: Ideally this should be a separate script or terraform. Then we can call that separate script or terraform from within this script.

gen3/bin/dbbackup.sh
```diff
@@ -43,19 +36,37 @@ vpc_name="$(gen3 api environment)"
...
+
+cluster_arn=$(kubectl config current-context)
+eks_cluster=$(echo "$cluster_arn" | awk -F'/' '{print $2}')
 
-gen3_log_info "policy_name: $policy_name"
 gen3_log_info "account_id: $account_id"
 gen3_log_info "vpc_name: $vpc_name"
 gen3_log_info "namespace: $namespace"
 gen3_log_info "sa_name: $sa_name"
 gen3_log_info "bucket_name: $bucket_name"
```
jawadqur, line 49: We should not need a unencrypted bucket,  we should only allow an encrypted bucket.

gen3/bin/dbbackup.sh
```diff
@@ -222,28 +224,289 @@ copy_to_aurora() {
...
+    gen3 job run psql-db-backup-encrypt
+}
+
+# Function to set up cronjob for encrypted backup
+setup_cronjob() {
+    gen3 job cron psql-db-backup-encrypt "15 1 * * *"
+}
+
+# Create policy for Mountpoint for Amazon S3 CSI driver
+create_s3_csi_policy() {
```
jawadqur, line 238: Same thing for all these other functions related to csi driver. These should go into a `kube-setup-s3-csi` or similar script, and then this script can call that other script, we need to make sure the other script is idempotent as well.

emalinowski (approved):

PR #2614, opened 2024-08-06
feat(set-es-domain): Updated es-proxy script to allow for setting es domain

emalinowski:
### New Features
Updated es-proxy script to allow for setting es domain


frankliuao:
Thank you Ed!

PR #2620, opened 2024-08-20
Update support emails from @datacommons.io to @gen3.org

smvgarcia:
update all support emails from <whatever-support>@datacommons.io to <whatever-support>@gen3.org
--do not merge until gen3.org emails are active.-- the emails are active now, so it can be merged whenever ready


jacob50231 (approved):
All looks good to me!

PR #2644, opened 2024-09-23
Feat/es proxy irsa no terragrunt

AidanHilt:
https://ctds-planx.atlassian.net/browse/GPE-1037


jawadqur (approved):
tested in qa-brh and it's working as expected!

PR #2645, opened 2024-09-23
add alerts to psql encrypted backup job

ajoaugustine:
Add alerts to postgresql backup job

### DIFF
--- a/kube/services/jobs/psql-db-backup-encrypt-job.yaml
+++ b/kube/services/jobs/psql-db-backup-encrypt-job.yaml
@@ -74,7 +74,6 @@ spec:
               ENV_NS_DIR="${HOST_DIR}/${environment}-${namespace}"
               DATE_DIR="${ENV_NS_DIR}/${date_str}"
               mkdir -p "${DATE_DIR}"
-              slack=false
               # PGP public key
               PUBLIC_KEY="-----BEGIN PGP PUBLIC KEY BLOCK-----
 
--- a/kube/services/jobs/psql-db-backup-encrypt-job.yaml
+++ b/kube/services/jobs/psql-db-backup-encrypt-job.yaml
@@ -143,16 +143,12 @@ spec:
               # Import the public key
               echo "$PUBLIC_KEY" | gpg --import
 
-              # Check and retrieve Slack webhook if not provided
-              slack=false
-              if [[ "${slackWebHook}" == 'None' || -z "${slackWebHook}" ]]; then
-                slackWebHook=$(g3kubectl get configmap global -o jsonpath={.data.slack_alarm_webhook})
-              fi
-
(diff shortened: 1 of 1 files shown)


kube/services/jobs/psql-db-backup-encrypt-job.yaml
```diff
@@ -206,6 +211,7 @@ spec:
               # Logging the failed backups
               if [ -n "$failed_backups" ]; then
                 gen3_log_info "Failed backups:\n$failed_backups"
+                curl -X POST --data-urlencode "payload={\"text\": \"*Backup failed* for psql-db-backup-encrypt on ${hostname}  Cluster: ${environment} Namespace: ${namespace} at $(date).\nFailed backups: ${failed_backups}\"}" $slackWebHook
```
jawadqur, line 214: Can we check if the slackwebhook is set to alert, and if it's not don't alert. Similar to other jobs in cloud-auto

kube/services/jobs/psql-db-backup-encrypt-job.yaml
```diff
@@ -138,6 +143,18 @@ spec:
               # Import the public key
               echo "$PUBLIC_KEY" | gpg --import
 
+              # Check and retrieve Slack webhook if not provided
+              slack=false
+              if [[ "${slackWebHook}" == 'None' || -z "${slackWebHook}" ]]; then
+                slackWebHook=$(g3kubectl get configmap global -o jsonpath={.data.slack_alarm_webhook})
+              fi
```
jawadqur, line 150: We should not do kubectl operations from within the job itself, we should mount the secrets/configmaps needed and in this script just check if they're present.

kube/services/jobs/psql-db-backup-encrypt-job.yaml
```diff
@@ -38,6 +38,11 @@ spec:
                   key: environment
             - name: JENKINS_HOME
               value: "devterm"
+            - name: slackWebHook
+              valueFrom:
+                configMapKeyRef:
+                  name: global
+                  key: slack_alarm_webhook
```
jawadqur, line 45: Can we add optional: true like here:

https://github.com/uc-cdis/cloud-automation/blob/ef9f921d93b49dafb1005ff8eda1226bd21fede4/kube/services/jobs/etl-job.yaml#L73-L78

kube/services/jobs/psql-db-backup-encrypt-job.yaml
ajoaugustine: done. thanks

kube/services/jobs/psql-db-backup-encrypt-job.yaml
ajoaugustine: added

jawadqur (approved):

PR #2682, opened 2024-12-10
Add changes to ensure azlinux and mcrypt images are working fine

nss10:
Add changes to fence related deploy yamls, job and cron-job yamls,  to provide backwards compatibility with fence images using azlinux and old docker file with `python3.9-buster-2.0.0` image

I do see some files in


kube/services/jobs/usersync-job.yaml
```diff
@@ -93,7 +93,7 @@ spec:
             name: "projects"
       containers:
       - name: usersync
-        GEN3_FENCE_IMAGE
+        GEN3_FENCE_MCRYPT_IMAGE
```
nss10, line 96: I think adding this, enforces all commons to have `fence_mcrypt` in their manifest and doesn't seem like a great idea

AlbertSnows (dismissed):
Approved assuming the pipeline issues are resolved. can you add a description to the PR so I or other reviewers have a better understanding of the purposes of these changes? The PR looks fine, but due to lack of context there may be things I am missing. Getting another review from someone else will also be beneficial.

kube/services/jobs/usersync-job.yaml
Avantol13: Yeah... It will force every job to use the new image for the usersync job regardless of if they're using the latest Fence, which is going to change what every Prod env runs without them explicitly updating a manifest or anything. We shouldn't do this. It doesn't allow someone to test the update in a staging/preprod before deploying in prod. We want to only use this new image if they're deploying the new Fence as configured in the manifest. Maybe we need a new manifest entry

nss10: I suggested adding a new `fence_mcrypt` image line in the manifest's versions, specifically for jobs where usersync requires this branch. The team would then need to manually update the usersync job in the AdminVM.

Need opinions.

jawadqur (dismissed):

jawadqur (approved):
lgtm! lots of old fun, we might have to think of manually updating existing cronjobs as I don't think this updates it, but they will eventually fail and we should see them bubble up (hopefully)

PR #2686, opened 2024-12-12
Fix Google Bucket Job

BinamB:
Link to JIRA ticket if there is one: https://ctds-planx.atlassian.net/browse/DCF-1947
Needed to add `gibberish-detector` to pre-commit since pre-commit tries to call it but it doesn't automatically install it. 

### New Features
- Add map_file option to validation
- Add GDC_poject_map mount on Google Replication

### Dependency updates
- Add `gibberish-detector` to pre-commit hook.


MaribelleHGomez (dismissed):
changes look good to me
I think we should add a note about why the mapping arg was added to the job for historical reference

jacob50231 (dismissed):

jacob50231 (approved):

PR #2695, opened 2025-01-13
(HP-1825): Script for automated launch test of workspaces

BinamB:

### New Features
- New job and cronjob to test launching workspaces.


kube/services/jobs/workspace-launch-test-cronjob.yaml
```diff
@@ -0,0 +1,197 @@
+# To run: gen3 job run workspace-launch-test-cronjob COMMONS_URL <commons base url> NOTEBOOK <NOTEBOOK name>
+# example: gen3 job run workspace-launch-test-cronjob COMMONS_URL https://qa-heal.planx-pla.net NOTEBOOK '(Generic) Jupyter Lab Notebook with R Kernel'
+# If NOTEBOOK isn't provided, it defaults to '(Generic) Jupyter Lab Notebook with R Kernel'
```
mfshao, line 3: So this means this cronjob can only test one selected workspace launch a time right? I would see if this fits @jawadqur 's need
I think it would be better if we can mount a list of workspace names into this job, and triggers job script over this list of workspace names. This might be easier for using it, but if it is significant harder we can also defer it to another ticket

BinamB: I see what you mean. I think it shouldn't be too difficult to change

files/scripts/workspaces_launch_test.py
```diff
@@ -0,0 +1,225 @@
...
+        try:
+            options_response = requests.get(options_url, headers=self.headers)
+            options_response.raise_for_status()
+
+        except requests.exceptions.RequestException as e:
+            error_msg = f"Couldn't get workspace options with error: {e}"
+            logging.error(error_msg)
+            self.reason_for_failure = error_msg
+        
+        options = options_response.json()
```
mfshao, line 88: nitpick: it might be better to convert `options` into a dict here with `name: id` pairs, and then create `available_images` and simplfies some logic in `get_info_for_image()`

mfshao (approved):
👍

mfshao (dismissed):

files/scripts/workspaces_launch_test.py
```diff
@@ -177,12 +177,13 @@ def start_workspace_launch_test(self, image_name, workspace_id):
...
         
 
-    def monitor_workspace_status(self, interval=10):
+    def monitor_workspace_status(self, interval=10, timeout_limit=10):
         """
         In an interval of given time (in seconds) hit the workspace status endpoint to monitor the status of the workspace
 
         Args:
             interval (int, optional): Interval (in seconds) to hit the options endpoint. Defaults to 10 seconds.
+            timeout_limit(int, optional): Time limit at which a workspace launch is considered a failure.
```
mfshao, line 186: minor: probably should mention this value is in minutes, not seconds

mfshao (approved):

PR #2708, opened 2025-01-24
Add `poetry run` for fence-create with backwards compatibility

nss10:
### New Features
* Increase support to amazon linux fence image with adding `poetry run fence-create`

### DIFF
--- a/gen3/bin/kube-setup-ohdsi.sh
+++ b/gen3/bin/kube-setup-ohdsi.sh
@@ -11,7 +11,13 @@ export namespace=$(gen3 api namespace)
 new_client() {
   atlas_hostname="atlas.${hostname}"
   gen3_log_info "kube-setup-ohdsi" "creating fence oidc client for $atlas_hostname"
-  local secrets=$(g3kubectl exec -c fence $(gen3 pod fence) -- fence-create client-create --client atlas --urls https://${atlas_hostname}/WebAPI/user/oauth/callback?client_name=OidcClient --username atlas --allowed-scopes openid profile email user | tail -1)
+  
+  # Adding a fallback to `poetry run fence-create` to cater to fence containers with amazon linux.
+  local secrets=$(
+    (g3kubectl exec -c fence $(gen3 pod fence) -- fence-create client-create --client atlas --urls https://${atlas_hostname}/WebAPI/user/oauth/callback?client_name=OidcClient --username atlas --allowed-scopes openid profile email user | tail -1) 2>/dev/null || \
+      g3kubectl exec -c fence $(gen3 pod fence) -- poetry run fence-create client-create --client atlas --urls https://${atlas_hostname}/WebAPI/user/oauth/callback?client_name=OidcClient --username atlas --allowed-scopes openid profile email user | tail -1
+
+  )
   # secrets looks like ('CLIENT_ID', 'CLIENT_SECRET')
   if [[ ! $secrets =~ (\'(.*)\', \'(.*)\') ]]; then
     gen3_log_err "kube-setup-ohdsi" "Failed generating oidc client for atlas: $secrets"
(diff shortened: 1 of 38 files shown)


kube/services/jobs/client-modify-job.yaml
```diff
@@ -125,11 +125,11 @@ spec:
 
             echo Starting to collect client list     
 
-            CLIENT_LIST=$(fence-create client-list | grep "'name'")
+            CLIENT_LIST=$((fence-create client-list || poetry run fence-create client-list) | grep "'name'")
             CLIENT_LIST=${CLIENT_LIST//"'name': "/}
             CLIENT_LIST=${CLIENT_LIST//"'"/}
             CLIENT_LIST=${CLIENT_LIST//","/}
-            CLIENT_LIST=($(echo $CLIENT_LIST | tr " "))
+            CLIENT_LIST=($(echo $CLIENT_LIST | xargs))
```
nss10, line 132: Replacing `tr` with `xargs` assuming we wanted to tokenize it

gen3/bin/kube-setup-ohdsi.sh
```diff
@@ -11,7 +11,13 @@ export namespace=$(gen3 api namespace)
 new_client() {
   atlas_hostname="atlas.${hostname}"
   gen3_log_info "kube-setup-ohdsi" "creating fence oidc client for $atlas_hostname"
-  local secrets=$(g3kubectl exec -c fence $(gen3 pod fence) -- fence-create client-create --client atlas --urls https://${atlas_hostname}/WebAPI/user/oauth/callback?client_name=OidcClient --username atlas --allowed-scopes openid profile email user | tail -1)
+  
+  # Adding a fallback to `poetry run fence-create` to cater to fence containers with amazon linux.
+  local secrets=$(
+    (g3kubectl exec -c fence $(gen3 pod fence) -- fence-create client-create --client atlas --urls https://${atlas_hostname}/WebAPI/user/oauth/callback?client_name=OidcClient --username atlas --allowed-scopes openid profile email user | tail -1) 2>/dev/null || \
```
AlbertSnows, line 17: these long command lines are a bit hard to read. if it's plausible I would suggest breaking these up into smaller parts and/or abstracting the behavior of some of strings if they are identical enough. Not required for approval, but worth mentioning.

kube/services/jobs/client-modify-job.yaml
AlbertSnows: I think this works, I noticed xargs has the -n command, which might be useful? 
example: https://explainshell.com/explain?cmd=echo+%24CLIENT_LIST+%7C+xargs+-n1

AlbertSnows:
Github is recommending changing line `129` of the `Dockerfile` to something like this:
`CMD ["/bin/bash"]`

AlbertSnows (approved):
looks like most of the changes are adding `poetry run` which should be fine, but I would suggest getting another review before merging.

kube/services/jobs/client-modify-job.yaml
nss10: Thanks for the comment, I'm a little unsure on how a `--max-args` field be useful in this case? 

I see the `tr` command failing in the original source code, and I was assuming the intention here is to split the list into space seprerated characters barring any amount of extra spaces, I think `xargs` itself does a pretty good job. 
Moreover, we don't have any control on number of elements we can have in `$CLIENT_LIST`, so I don't full understand the use of `-n` here.

gen3/bin/kube-setup-ohdsi.sh
nss10: For this PR in particular, The main goal of this PR is to add a fallback to the `poetry run fence-create` to support the amazon linux images. I don't think it is wise to get into the nitty gritty of the commands at this time, in this PR. 

Thanks for pointing this one out though. I do think cleaning this up would improve maintainability.

kube/services/jobs/client-modify-job.yaml
AlbertSnows: actually yeah nevermind, we're just storing it in client list so i think adding max args would be pointless

jawadqur (approved):

PR #2716, opened 2025-01-31
Revert/fallback_to_poetry_run

nss10:
Reverting changes from 
* Add changes to ensure azlinux and mcrypt images are working fine (#2682)
* Add poetry run for fence-create with backwards compatibility (#2708)

Note: These changes were originally made to fallback to `poetry run python` or `poetry run fence-create` to support the new AL2-based fence image. However, updating the `PATH` variable in the Docker image allows us to avoid all this extra work.

This PR needs to be tested with both the current fence:master and the chore/ccrypt_usersync branch, as the latter uses Amazon Linux. We want to ensure that the changes in this cloud-auto branch work for both versions without causing any breakages.

### DIFF
--- a/.github/workflows/integration_tests.yaml
+++ b/.github/workflows/integration_tests.yaml
@@ -12,6 +12,7 @@ jobs:
     uses: uc-cdis/.github/.github/workflows/integration_tests.yaml@master
     with:
       QUAY_REPO: "awshelper"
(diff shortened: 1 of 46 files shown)


krishnaa05 (approved):
lgtm -> Test cases have passed

PR #2718, opened 2025-02-05
Usersync fixes for ccrypt

BinamB:
### Improvements
We are releasing a new Amazon Linux 2023 image that is more secure. In order to do this we need to patch Paramiko so md5 is not blocked by fips. We also needed to remove some of the key exchange algorithms from the ssh config. (backwards compatible)


nss10 (approved):
LGTM!

PR #2722, opened 2025-02-06
Update deprecated js_include in nginx configs

jawadqur:

### Improvements
- Update nginx config to remove deprecated js_include (It was deprecated quite a while ago, in 0.4.0 or something)

### DIFF
--- a/kube/services/revproxy/nginx.conf
+++ b/kube/services/revproxy/nginx.conf
@@ -247,7 +247,7 @@ server {
   more_set_headers "Access-Control-Expose-Headers: Content-Length,Content-Range";
 
   # update service release cookie
-  add_header Set-Cookie "service_releases=${service_releases};Path=/;Max-Age=600;HttpOnly;Secure;SameSite=Lax";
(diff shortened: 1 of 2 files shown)


kube/services/revproxy/nginx.conf
```diff
@@ -247,7 +247,7 @@ server {
   more_set_headers "Access-Control-Expose-Headers: Content-Length,Content-Range";
 
   # update service release cookie
-  add_header Set-Cookie "service_releases=${service_releases};Path=/;Max-Age=600;HttpOnly;Secure;SameSite=Lax";
+  # add_header Set-Cookie "service_releases=${service_releases};Path=/;Max-Age=600;HttpOnly;Secure;SameSite=Lax";
```
mfshao, line 250: is this intentional? I thought this is to add some madatory security related cookies

jawadqur: This is/ was used for some canary deployments, which was never fully implemented. We could keep it in, but I don't think we need it.

mfshao (approved):

PR #2724, opened 2025-02-07
adding commands to allow for the new ccrypt image to work (backwards …

EliseCastle23:
…compatible)

### Improvements
We are releasing a new Amazon Linux 2023 image that is more secure. In order to do this we need to patch Paramiko so md5 is not blocked by fips. We also needed to remove some of the key exchange algorithms from the ssh config.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -2396,57 +2396,6 @@
         "line_number": 44
       }
     ],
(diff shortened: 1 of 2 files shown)


EliseCastle23:
https://github.com/uc-cdis/cloud-automation/pull/2718

PR #2740, opened 2025-02-24
Update dicom-server-deploy.yaml

jawadqur:

### DIFF
--- a/kube/services/dicom-server/dicom-server-deploy.yaml
+++ b/kube/services/dicom-server/dicom-server-deploy.yaml
@@ -14,6 +14,7 @@ spec:
       labels:
         app: dicom-server
         release: production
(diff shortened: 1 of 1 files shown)


paulineribeyre (approved):
dicom-server was working in MIDRC staging because this was manually added to the deployment. Without it, network policies prevent dicom-server from connecting to the "orthanc" DB.

Things were working in MIDRC prod because there existed a "dicom-server" g3auto secret containing the same DB creds as the "orthanc" secret, and network policies are created based on those secrets (netpol assume the dicom-server service uses the "dicom-server" secret). When the DB creds were changed, but the "unused" "dicom-server" secret was not updated, things stopped working.

This change should allow us to skip setting up network policies for this service.

PR #2762, opened 2025-04-01
Add prometheus metrics support to gen3-workflow

nss10:
Link to JIRA ticket if there is one: [MIDRC-857](https://ctds-planx.atlassian.net/browse/MIDRC-857)

### New Features
* Add prometheus metrics support to gen3-workflow

[MIDRC-857]: https://ctds-planx.atlassian.net/browse/MIDRC-857?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/kube/services/gen3-workflow/gen3-workflow-deploy.yaml
+++ b/kube/services/gen3-workflow/gen3-workflow-deploy.yaml
@@ -16,6 +16,9 @@ spec:
       maxUnavailable: 0
   template:
     metadata:
+      annotations:
+        prometheus.io/scrape: "true"
+        prometheus.io/path: /metrics/
       labels:
         app: gen3-workflow
         release: production
--- a/kube/services/gen3-workflow/gen3-workflow-deploy.yaml
+++ b/kube/services/gen3-workflow/gen3-workflow-deploy.yaml
@@ -96,3 +96,12 @@ spec:
             memory: 1Gi
           limits:
             memory: 1Gi
+        command: ["/bin/sh", "-c"]
+        args:
+        - |
+          #!/bin/sh
+          # Set the prometheus multiproc dir to the env var
+          export PROMETHEUS_MULTIPROC_DIR=$(grep 'PROMETHEUS_MULTIPROC_DIR:' /src/gen3-workflow-config.yaml | awk -F': ' '{print $2}' | tr -d '"')
+          echo "PROMETHEUS_MULTIPROC_DIR is set to - $PROMETHEUS_MULTIPROC_DIR"
+          # Run the gen3-workflow server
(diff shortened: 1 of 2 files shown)


kube/services/gen3-workflow/gen3-workflow-deploy.yaml
```diff
@@ -96,3 +96,12 @@ spec:
             memory: 1Gi
           limits:
             memory: 1Gi
+        command: ["/bin/sh", "-c"]
+        args:
+        - |
+          #!/bin/sh
+          # Set the prometheus multiproc dir to the env var
+          export PROMETHEUS_MULTIPROC_DIR=$(grep 'PROMETHEUS_MULTIPROC_DIR:' /src/gen3-workflow-config.yaml | awk -F': ' '{print $2}' | tr -d '"')
```
nss10, line 104: For prometheus to create metrics in multiprocess mode, it is important that the `PROMETHEUS_MULTIPROC_DIR` is set as an env before the process starts. [Link here](https://prometheus.github.io/client_python/multiprocess/#:~:text=Prepend%20’live’%20to%20the%20beginning%20of%20the%20mode%20to%20return%20the%20same%20result%20but%20only%20considering%20living%20processes(e.g.%2C%20’liveall%2C%20’livesum’%2C%20’livemax’%2C%20’livemin’%2C%20’livemostrecent’).)

nss10: This step is already mentioned in `bin/setup_prometheus` in the gen3-workflow service[here](https://github.com/uc-cdis/gen3-workflow/blob/2a1d4b860278b37196ed157a01c12f55b282057e/bin/setup_prometheus#L33), but was never called during service setup. If we manage to run the script during `bin/_common_setup.sh` as a part of launching the service, we no longer need to add additional logic here to set the env var.

paulineribeyre: How come we don't need to do this for gen3-user-data-library and fence?
i think it'd be better do it in the service code if we can 🤔

nss10: This logic is now placed in a shell script called [_common_setup.sh](https://github.com/uc-cdis/gen3-workflow/blob/d16fba853077608ca667103521e62dca013a0e14/bin/_common_setup.sh#L9-L16), which is sourced by run.sh. The run.sh script is executed as part of the Dockerfile’s `CMD` instruction.

paulineribeyre (approved):

paulineribeyre:
please update the PR title

PR #2811, opened 2025-06-30
Reduce WTS liveness interval to 20 seconds

george42-ctds:
JIRA ticket: [HP-2177](https://ctds-planx.atlassian.net/browse/HP-2177) 

### Improvements

* Reduce WTS liveness interval to 20 seconds

### DIFF
--- a/kube/services/wts/wts-deploy.yaml
+++ b/kube/services/wts/wts-deploy.yaml
@@ -98,7 +98,7 @@ spec:
             port: 80
           initialDelaySeconds: 30
           periodSeconds: 60
-          timeoutSeconds: 30
+          timeoutSeconds: 20
           failureThreshold: 10
         readinessProbe:
           httpGet:


kube/services/wts/wts-deploy.yaml
```diff
@@ -97,7 +97,7 @@ spec:
             path: /_status
             port: 80
           initialDelaySeconds: 30
-          periodSeconds: 60
+          periodSeconds: 20
```
mfshao, line 100: I think `livenessProbe` is actuaaly good, but it is `readinessProbe` taht are pinging WTS every 10 seconds, and we should change that probe to 20 or 30 seconds

george42-ctds: Great, thank you! The readinessProbe has been added.

Should the livenessProbe get reverted to 60 seconds?

mfshao: yeah lets put liveness probe back to 60

mfshao (approved):

PR #2885, opened 2026-08-27
fix(hatchery): if it can create roles it ought to be able to get what…

jbarno:
From discussion. Spelunking through aws I found this list:
https://cdis.slack.com/archives/CLZJVC38B/p1787857535895259

brh-staging service account has been manually updated, we'll likely need to do the same for prod when we swap.


jawadqur:
We're no longer using cloud-automation.

